PR #1, opened 2021-11-12
Fix pulsed labels

timoML:
Fixes the x axis of exported pulsed thumbnails from html tags (`<sub>`). 

## Description
While the pulsed gui (based on pyqtgraph) supports html tags for the labels, the thumbnail export via matplotlib does not.
This fix transforms (only) html superscript `<sub>` tags to the required Latex format. Currently, `<sub>` is the only html tag used in predefined methods.

## How Has This Been Tested?
Dummy mode.

Buggy:
image
Fixed:
image

## Types of changes
- [x] Bug fix
- [ ] New feature

## Checklist:


NicolasStaudenmaier (approved):
Looks good and works fine. Thanks.

PR #3, opened 2021-11-12
Fast waveform upload on Keysight awg819x 

timoML:
Using the IVI-COM driver from Keysight, waveform upload speeds can be massively boosted (>200 MB/s via PCIe).

## Description
Caveats:
- due to comtypes usage, only works on Windows
- the limiting factor in total upload time seems now often the binary sample compilation needed after getting floats from the sequence generator

## How Has This Been Tested?
- 3 setups with both awg8190, 95.

## Types of changes
- [ ] Bug fix
- [x] New feature

## Checklist:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,6 +32,7 @@ windows_dep = [
     'PySide2>=5.15.2',
     'PyVisa>=1.11.3',
     'scipy>=1.7.1',
+    'comtypes>=1.1.8'
 ]
 
 with open('VERSION', 'r') as file:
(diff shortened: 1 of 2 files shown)


timoML:
Please note that there is a confirmed memory leak occuring somewhere between the Keysight driver and comtypes (1.2.0). Uploading multiple waves will cause memory usage rise without proper clearing. Thanks @PJVetter for investigating.

TobiasSpohn:
Requires additional research to check it's improvement

PR #6, opened 2021-12-01
Bug fix to ODMR scan power

kay-jahnke:
ODMR scan power was just not handed through from GUI to the hardware file.

## Description

The gui element was declared but never connected to anything. So the fix was to simply hand through the value from the gui to the logic and it worked.

## Motivation and Context

Bug Fix.

## How Has This Been Tested?

Tested on Win10 21H1 with pure python 3.8.10 on dummy hardware. The debug message now shows the correct value.

## Types of changes

- [x] Bug fix
- [ ] New feature

## Checklist:


christoph-mueller (approved):
tested on LP1 and works as intended

PR #7, opened 2021-12-08
[Fix] Pulse editor gui 

timoML:
## Description
The pulse editor gui was not showing digital channels correctly in its tabular view. The fix reverts changes made during the port; potentially by mistake.

Working again:
image

## Types of changes
- [x] Bug fix
- [ ] New feature

## Checklist:


timoML:
@Neverhorst: I don't understand the reasoning why the self.model().laserRole was removed in the first place. Do you recall? Otherwise, seems to work again in dummy mode for me.

kay-jahnke (approved):
Tested on dummy and with hardware on LP1: works fine, THANKS

PR #8, opened 2022-01-18
bug fix to qdplotter and benchmarking tool

kay-jahnke:
For some reason numpy can't do a `np.isfinite()` on its own data-types. Therefore we convert it to list first.

The benchmark called `self.log` although it wasn't a qudi module, causing an error. Replaced that by a `print`.

## Description

A bug in qdplotter showed up, if only one point was plotted and the x-axis is a ndarray.

Errors in stand-alone modules need to be printed and can't be posted via `self.log.error`

## Motivation and Context

Fixing bugs.

## How Has This Been Tested?

Tested with current core on NVision/LP1 running on Win 10 21H1

## Types of changes

- [x] Bug fix
- [ ] New feature

## Checklist:

### DIFF
--- a/src/qudi/gui/qdplot/qdplot_gui.py
+++ b/src/qudi/gui/qdplot/qdplot_gui.py
@@ -389,7 +389,7 @@ class QDPlotterGui(GuiBase):
                 symbol='d',
                 symbolSize=6,
                 symbolBrush=mkColor(pen_color)))
-            self._plot_curves[plot_index][-1].setData(x=xd, y=yd)
+            self._plot_curves[plot_index][-1].setData(x=list(xd), y=yd)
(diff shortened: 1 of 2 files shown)


christoph-mueller (approved):
tested and works as intended

Neverhorst:
@kay-jahnke @christoph-mueller Can someone please elaborate? I don't get any errors when plotting arrays of len == 1. Also what do you mean with `np.isfinite`? Typecasts on numpy arrays can be costly, so I would want to avoid that if posssible.

Also errors in non-logic/hardware/gui modules should not be logged using `print()` but via the Python `logging` module (see [documentation](https://ulm-iqo.github.io/qudi-core/design_concepts/logging.html) under "other modules").
In 99% of the cases you should not log any errors from standalone modules anyway but raise an exception!

Please don't make "pro-forma PRs"

PR #9, opened 2022-01-26
fix: not-fulfilled switch interface

timoML:
PulseBlaster was broken before due to not fulfilled SwitchInterface.
Fixed with dummy implemntations.

## How Has This Been Tested?
On real setup (M26/4219)

## Types of changes
- [x] Bug fix
- [ ] New feature

## Checklist:

### DIFF
--- a/src/qudi/hardware/spincore/pulse_blaster_esrpro.py
+++ b/src/qudi/hardware/spincore/pulse_blaster_esrpro.py
@@ -174,7 +174,7 @@ class PulseBlasterESRPRO(SwitchInterface, PulserInterface):
                    16: 'Scanning'}
 
     # For switch interface:
-    switch_states = {'d_ch1': False, 'd_ch2': False, 'd_ch3': False,
+    _switch_states = {'d_ch1': False, 'd_ch2': False, 'd_ch3': False,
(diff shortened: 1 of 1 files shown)


kay-jahnke (changes requested):
In my opinion, either the hardware implements the interface (it inherits the interface class) or it does not implement it and therefore also does not inherit it. It therefore does not make sense to pretend to implement it and then raise.

Hence I would suggest either to really implement the switch interface or completely drop it from the inheritance.

timoML:
I agree. Since I don't know anyone using the SwitchInterface, I went for the latter.

kay-jahnke:
A valid decision. 
But the import is still in there, plus the methods for the old switch interface (line 176ff and 1090ff). please kill all of it, as it is non-functional.

Have you tested that the rest still works?

timoML:
Ok, I didn't see those other methods. I used them to implement in the SwitchInterface now.
Basic tests with simple TTL sequences from the pulsed gui were done on Ilia's  setup after getting the file to run again.
He will continue to work with it; so maybe it's worth waiting for his feedback before merging...

Neverhorst (approved):

kay-jahnke (approved):
can't test it on hardware, but the code looks much better now.

PR #11, opened 2022-02-07
Bug fix spectrometer

kay-jahnke:
In the continuous mode for background acquisition, the signature of the signal did not match the signature of the method. Fixed that.

Also added a maximum repetitions when using continuous acquisition.

## Motivation and Context

Continuous acquisition would literally run forever. There are however cases, where you would like to limit that to a maximum. If you do not want a limit on the acquisitions, just leave the max at 0.

Max repetitions is a status variable of the logic and can be set via the settings dialog in the GUI.

## How Has This Been Tested?

Tested with dummy module on Win10 Pro 21H2.

## Types of changes

- [x] Bug fix
- [ ] New feature

## Checklist:

### DIFF
--- a/src/qudi/gui/spectrometer/settingsdialog.py
+++ b/src/qudi/gui/spectrometer/settingsdialog.py
@@ -21,7 +21,7 @@ If not, see <https://www.gnu.org/licenses/>.
 """
 
 from PySide2 import QtCore, QtWidgets
-from qudi.util.widgets.scientific_spinbox import ScienDSpinBox
+from qudi.util.widgets.scientific_spinbox import ScienDSpinBox, ScienSpinBox
 from qudi.util.widgets.toggle_switch import ToggleSwitch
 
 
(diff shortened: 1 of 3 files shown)


Maabur:
ok thanks Kay, I will have a look in the next few days

Maabur:
It seems fine to me I have tested it on windows 10.
I am just not familiar with the syntax using backslash in line 169  in the logic file. It is hard for me to understand what the if statement does. 
I know it works coz I tested it but could you you just explain it to me ? I tried to google it and could not get anything useful.
cheers

Maabur (approved):

kay-jahnke:
@Maabur the backslash at the end of a line essentially says, that the next line should be treated as if it was in the same line. So it is a line break without breaking the expression. For long expressions, this is a very essential feature.

See PEP8: https://www.python.org/dev/peps/pep-0008/

PR #14, opened 2022-03-09
Millennia ev laser

takuya-ulm:
## Description

- Bug fix
- Introduce VISA class and Command classes

## Motivation and Context

- Some bugs were found in transition to the new core
- Even though the visa package was used in several laser hardware, it was not encapsulated. 
## How Has This Been Tested?

Tested with Millennia eV laser.
This hardware change does not have a big effect on the other area of the code.

## Types of changes

- Bug fix
- ~~New feature~~
- ~~Breaking change (Causes existing functionality to not work as expected)~~

## Checklist:


TobiasSpohn (approved):
Thanks

PR #20, opened 2022-06-26
Timeseries restore gui view

timoML:
Restores the view settings of the timeseries gui.

## Description
Previously, the timeseries gui forgot about the channels set in its 'view settings' on reload of the module. Thus, all configured traces of the logic would show up. This PR uses a StatusVar to restore these settings plus the value of the "current value" text box.

## How Has This Been Tested?
in dummy

## Types of changes
- [ ] Bug fix
- [x] New feature

## Checklist:


src/qudi/gui/time_series/time_series_gui.py
```diff
@@ -299,6 +304,20 @@ def on_deactivate(self):
 
         self._mw.close()
 
+    @_view_settings.representer
+    def __repr_view_settings(self, value):
```
LukePiewalker, line 308: I've seen this a few times now in some other code, yet I don't really know why the core requires the representer to have "value" here. In most cases it is not used. Can you enlighten me about this @timoML or @Neverhorst?

src/qudi/gui/time_series/time_series_gui.py
```diff
@@ -299,6 +304,20 @@ def on_deactivate(self):
...
+
+        view_settings = {}
+
+        vis_settings = self.visible_settings
+        curr_ch = self._mw.curr_value_comboBox.currentText()
+        curr_ch_idx = self._mw.curr_value_comboBox.findText(curr_ch)
+
+        view_settings['current_channel_idx'] = curr_ch_idx
```
LukePiewalker, line 316: I think it would make more sense, to serialize channel string here instead of the index since upon change of the channels in the hardware, the gui would display a different channel on reload of the module while I think it should go to None in this case.

src/qudi/gui/time_series/time_series_gui.py
```diff
@@ -354,6 +373,15 @@ def _init_channel_settings_dialog(self):
         layout.setRowStretch(i + 1, 1)
         self._csd.trace_selection_scrollArea.setLayout(layout)
 
+    def _init_gui_view(self):
+        self.visible_settings = self._view_settings
+
+        # Set 'current channel' text box
+        try:
+            self.current_value_channel = self._view_settings['current_channel_idx']
```
LukePiewalker, line 381: See above, maybe store the channel string in the StatusVar and not the index.

LukePiewalker (changes requested):
Looks good to me, except that a reload of the module after the hardware channels have changed, would display the wrong value in the gui. (See comments in the code)

src/qudi/gui/time_series/time_series_gui.py
LukePiewalker: Should be resolved with 77cd1120dd4a7a65f8255fbb24b1e0e59a087a80

src/qudi/gui/time_series/time_series_gui.py
LukePiewalker: Should be resolved with 77cd1120dd4a7a65f8255fbb24b1e0e59a087a80

LukePiewalker (approved):
Did the requested changes myself.

src/qudi/gui/time_series/time_series_gui.py
Neverhorst: `StatusVar` will pass the currently stored object in the instance variable to the representer. If your status variable at runtime is for example a custom object for which the type can not be stored in a YAML file, you can use the representer to convert this object to a simple type. This could look like e.g.:
```python
class MyCustomObject:
    def __init__(a, b, c):
        self.a = a
        self.b = b
        self.c = c


class MyModule(LogicBase):
    
    _my_status_var = StatusVar(name='my_status_var', default=list())
        
    @_my_status_var.representer
    def _repr_my_status_var(self, value):
        return [value.a, value.b, value.c]
        
    @_my_status_var.constructor
    def _constr_my_status_var(self, value):
        return MyCustomObject(*value)
```

In fact getting the value of the `StatusVar` from somewhere in the module (i.e. from `self`) is actually bad design since the value of the `StatusVar` at runtime does not correspond to the one being dumped by the representer in that case. But it's possible...

Neverhorst (approved):
Thanks guys, looks good. 
The use of `StatusVar` is a bit misleading here since the GUI does not use its value at runtime to represent the current state but for now that's fine. In general the module could use a bit cleanup since it has some "organic growth" in it. Maybe I will have a look soon.
Let's merge this.

PR #29, opened 2022-08-31
PID toolchain rework

qku:
## Description
This PR makes the PID toolchain (`pidgui`, `pid_logic` and `software_pid_controller`) ready for `qudi-core`.

In addition, it adds units to setpoint, control and process variables. For this, additional methods for the `pid_controller_interface` were neccesary. This is a potentially breaking change as far as I understand, but should be easy to implement for existing hardware modules.

A normalization of the process value to the setpoint is now available as a config option. Very small (large) measurement values (e.g. nano Watts) would otherwise require very large (small) kP, kI, kD values.

The toolchain is further improved by clearly distinguishing the state of the recording loop and the state of the PID controller in the logic.

For testing, an option was added to `process_control_dummy` that enables a process value to change according to a set point, with some optional noise. This can be used to test the PID toolchain.

## Motivation and Context
The modules related to PID control were not compatible with the new core yet. We were working on a laser power stabilization setup based on a waveplate in a motorized rotation stage and a polarizing beam splitter cube. This now works with qudi.

## How Has This Been Tested?

With custom hardware modules for a powermeter, the rotation stage, and a custom interfuse from stage to a process setpoint device. It can also be tested with the process_control_dummy hardware.

## Screenshots:
pid_gui

## Types of changes
- [X] Bug fix
- [X] New feature

## Checklist:


timoML:
Thank you for the contribution. @GRexBayer can you test/have a look?

Our long term goal was to unify the PIDControllerInterface with the [ProcessSetpointInterface](https://github.com/Ulm-IQO/qudi-iqo-modules/blob/5e307c2f458cbd9afb3f2b4a995264acbd90f052/src/qudi/interface/process_control_interface.py#L105). It would be awesome if @qku could already look into that...

GRexBayer:
> Thank you for the contribution. @GRexBayer can you test/have a look?
> 
> Our long term goal was to unify the PIDControllerInterface with the [ProcessSetpointInterface](https://github.com/Ulm-IQO/qudi-iqo-modules/blob/5e307c2f458cbd9afb3f2b4a995264acbd90f052/src/qudi/interface/process_control_interface.py#L105). It would be awesome if @qku could already look into that...

Hi Timo,

Sorry for the late reply, sure I will have a look!

GRexBayer:
Hi,
As before, sorry for the late reply. 
First of all, thank you for the nice rework and effort put into this module!
I got the "dummy" version running and it looks good. At least it responding to my inputs and seems to be running stable.
From what I can tell the control values also behave the way they should.
The only issue I have, is that the example configs are not working properly. In one case it is mentioned in the file itself that the config interprets the value wrongly (dtype: float). In the other "speed" and "power" are not accepted values.
I guess this is also mentioned in the pull request regarding the interface file. We (QuDi team) can discuss it in our next meeting. I think this might not be a breaking change to the file, but who knows.

qku:
I also added a button to the GUI that resets the view, i.e. deletes the data in the plot. This is very useful if e.g. the setpoint was changed by a large amount. Then small deviations between setpoint and process value are not visible anymore - resetting the view deletes the old data and zooms in again.
image

qku:
According to my tests, this should work now with the changes to `process_control_interface` made in #37. I added the PID modules to `default.cfg` and extended `process_control_dummy` with an option where a process value changes according to a setpoint. This allows for easy testing of the PID toolchain with the dummy modules.

qku:
@GRexBayer thank you for your testing efforts so far. I believe that testing with real hardware is not strictly neccesary to verify the functionality of these changes. To facilitate testing with dummy modules, I added extra dependence functionality as outlined in my last comment.

GRexBayer:
> @GRexBayer thank you for your testing efforts so far. I believe that testing with real hardware is not strictly neccesary to verify the functionality of these changes. To facilitate testing with dummy modules, I added extra dependence functionality as outlined in my last comment.

You are right, Also in the Meeting this point was brought up. I can observe for the dummy as well that the voltage behaves roughly as I would expect. Maybe give me one more day to have a look at it, then I might merge it. From what I see, it works fine now

GRexBayer (dismissed):
This module was tested with the PM hardware module and with a dummy module. Is working fine after a few adjustments.

GRexBayer (approved):
A conflict in the docs had to be resolved.

PR #31, opened 2022-09-14
add a new hardware file

liuyanulm:
add a hardware file to control the Keysight power supply.

Using the keysight power supply for power switches, so that one can control devices such as APD, MINI amplifers etc remotely.
## Description

coded the hardware file of the Keysight power supply in the format of `switches`, it works pretty well.
## Motivation and Context

With this change, you can switch on your APDs and amplifiers and other devices remotely.
## How Has This Been Tested?

it was tested on the latest `iqo-qudi-modules` on a Windows 10 PC with a Keysight E36234A power supply.

image

## Types of changes

- [ ] Bug fix
- [x] New feature

## Checklist:

### DIFF
--- a/src/qudi/hardware/switches/power_switch_keysight.py
+++ b/src/qudi/hardware/switches/power_switch_keysight.py
@@ -1,0 +1,181 @@
+# -*- coding: utf-8 -*-
+
+"""
+Control the Radiant Dyes flip mirror driver through the serial interface.
+
+Copyright (c) 2021, the qudi developers. See the AUTHORS.md file at the top-level directory of this
+distribution and on <https://github.com/Ulm-IQO/qudi-iqo-modules/>
+
+This file is part of qudi.
+
+Qudi is free software: you can redistribute it and/or modify it under the terms of
+the GNU Lesser General Public License as published by the Free Software Foundation,
+either version 3 of the License, or (at your option) any later version.
+
+Qudi is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+See the GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License along with qudi.
+If not, see <https://www.gnu.org/licenses/>.
+"""
+
+try:
+    import visa
+except ImportError:
+    import pyvisa as visa
+from qudi.core.configoption import ConfigOption
+from qudi.core.statusvariable import StatusVar
+from qudi.util.mutex import RecursiveMutex
+from qudi.interface.switch_interface import SwitchInterface
+
+
+class PowerSwitch(SwitchInterface):
+    """ This class is implements communication with the Keysight E36234A power supply using pyVISA
+
+    Example config for copy-paste:
+
+    power_switch3:
+        module.Class: 'switches.power_switch_keysight.PowerSwitch'
+        options:
+            interface: 'USB0::0x2A8D::0x3402::MY59001224::INSTR'
+            hardware_name: 'Keysight_Power2'  # optional
+            switch_time: 1  # optional
+            remember_states: False  # optional
+            switches:
+                APD: ['OFF', 'ON']  # optional
+                Amplifier: ['OFF', 'ON']  # optional
+    """
+
+    # customize available switches in config. Each switch needs a tuple of at least 2 state names.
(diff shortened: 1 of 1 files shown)


src/qudi/hardware/switches/power_switch_keysight.py
```diff
@@ -0,0 +1,181 @@
...
+        """ Prepare module, connect to hardware.
+        """
+
+        self._resource_manager = visa.ResourceManager()
+        self._instrument = self._resource_manager.open_resource(self.serial_interface)
+        # set the maximum/protection output Voltage/current of the power supply
+        # Excelitas APD: 5V,1.3A
+        self._instrument.write('INST CH1')
+        self._instrument.write('OUTP 0')
+        self._instrument.write('VOLT 5')
```
timoML, line 82: Maximum voltages should be ConfigOptions, not hardcoded.

src/qudi/hardware/switches/power_switch_keysight.py
```diff
@@ -0,0 +1,181 @@
...
+    # name of the serial interface where the hardware is connected.
+    # Use e.g. the Keysight IO connections expert to find the device.
+    serial_interface = ConfigOption('interface', 'USB0::0x2A8D::0x3402::MY59001224::INSTR')
+
+    # StatusVariable for remembering the last state of the hardware
+    _states = StatusVar(name='states', default=None)
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.lock = RecursiveMutex()
```
timoML, line 67: I think the lock is not needed in this module.

timoML:
@liuyanulm Is the power supply API compatible with the existing hw file for the ([E3631A](url))[https://github.com/Ulm-IQO/qudi-iqo-modules/blob/main/src/qudi/hardware/power_supply/Keysight_E3631A.py]?

liuyanulm:
> 

Interesting idea. I also had a similar thought when I notice the `power_supply` hardware pacage. But after I went through the `E3631A` hardware file, I noticed that is not the category of functions I had in mind. fullfilling both the ProcessControlInterface and the SwitchInterface seems rather an ideal arrangement but I assume it's not easy. But we can keep this in mind.

src/qudi/hardware/switches/power_switch_keysight.py
liuyanulm: Now the voltages and currents are taken from `ConfigOption`

src/qudi/hardware/switches/power_switch_keysight.py
liuyanulm: These lines are used in all `switch hardwares`. I think we may keep it for now for unification

src/qudi/hardware/switches/power_switch_keysight.py
```diff
@@ -0,0 +1,181 @@
+# -*- coding: utf-8 -*-
+
+"""
+Control the Radiant Dyes flip mirror driver through the serial interface.
```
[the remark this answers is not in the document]
liuyanulm: corrected

liuyanulm:
> 

The Keysight power supplies are controlled not via API, you don't have to install any APP in your control computer. They are controlled via SCPI commands through USB connections, similar to how the SMIQ MW are controlled. The SCPI language of the devices are mostly the same.

liuyanulm:
> @liuyanulm Is the power supply API compatible with the existing hw file for the [E3631A](https://github.com/Ulm-IQO/qudi-iqo-modules/blob/main/src/qudi/hardware/power_supply/Keysight_E3631A.py)? In this case, we might think of unifying the files and fullfill both the ProcessControlInterface and the SwitchInterface.

@timoML The Keysight power supplies are not controlled via API, you don't have to install any APP in your control comuter. They are controlled via SCPI commands through USB connections. The range of the outputs are of course different.

timoML:
We discussed the PR in the maintenance team. Our prefered solution is to generalize your code such that every device fullfilling the ProcessControlInterface can fullfill the SwitchInterface. We will look into this and keep this PR pending meanwhile.

liuyanulm:
> 

Thanks

LukePiewalker:
Dear @liuyanulm ,

Since there now is the switch functionality for all ProcessControlInterface Hardware through PR #37 could you maybe incoorporate these changes in your PR?

timoML:
Closed for inactivity. Please re-open if you plan to make Keysight E36234A compatible with #37.

PR #33, opened 2022-09-20
Fix ni scanner deadlock

timoML:
Fixes a deadlock of the optimizer that causes the ni_scanning_probe_interfuse to hang.

## Description
Currently, there is a rare chance that the confocal gui (in more detail the optimizer) deadlocks when a cursor movement is executed right before/after a hardware timed scan. Most of the times this is observed after a 2D optimization with subsequent move to the optimized position. Then, the following 1D scan would not execute and the ni_scanner not return to unlocked module state.
This fix takes more care in protecting the shared data (scan arrays & cursor_write_queue) with 2 seperate (non-recursive) mutexes.  
In the future, the ni_scanning_probe_interfuse needs more rework. Currently, it's fetching data from hardware on calls to get_scan_data() from the logic. This should be decoupled and the ni_scanning_probe_interfuse should implement it's own polling loop.

## How Has This Been Tested?
setup2 (@anjusha-vs), @NicolasStaudenmaier setup, @LukePiewalker setup. Runs without deadlock for > 30 mins each.

## Types of changes
- [x] Bug fix
- [ ] New feature

## Checklist:


timoML:
There are some more commits from omniscan_dev in here (by @NicolasStaudenmaier fixing the poimanager). I think this is stable enough to merge together with this PR.

Neverhorst (approved):

PR #36, opened 2022-09-30
Qdplot improvements

Neverhorst:
## Description
Complete QDPlot overhaul. Makes use of new `qudi-core` module `qudi.util.widgets.plotting.interactive_curve`.
Improved stability.

`QDPlotLogic` has changed its access methods, so this is a breaking change for existing functionality that interfaces with qdplot.

## How Has This Been Tested?
On dummy cfg using console to set data and call all logic interface methods.
Win 10 21H2 (19044.2006)

## Screenshots:
image

## Types of changes
- [ ] Bug fix
- [x] New feature

## Checklist:


kay-jahnke (dismissed):
Tested the NVision evaporation logic on dummy and it works beautifully.

dummy on python 3.8, (old installation), Win 10 Pro 21H2

PR #37, opened 2022-09-30
Process control switch mixin

Neverhorst:
## Description
Added new mixin abstract class `qudi.interface.mixins.process_control_switch.ProcessControlSwitchMixin` for hardware modules implementing `ProcessControlInterface`, `ProcessSetpointInterface` or `ProcessValueInterface`.
Provides default implementations satisfying the `SwitchInterface` when mixed into an implementation of the above mentioned process control interfaces so it can be accessed as process control device _and_ as switch to toggle the device on/off.

Usage is simple... just inherit the mixin alongside `ProcessControlInterface`, `ProcessSetpointInterface` or `ProcessValueInterface`:
```python
class MyHardwareModule(ProcessControlSwitchMixin, ProcessControlInterface):
    ...
```

An implementation example is showcased using the `dummy.cfg` and loading the switch GUI. The dummy module `qudi.hardware.dummy.process_control_dummy.ProcessControlDummy` has been expanded using the new mixin to additionally satisfy the `SwitchInterface`.

> :warning: **Switching individual channels is not possible. Only enabling/disabling the entire device/all channels**

## Motivation and Context
The interfaces contained in `qudi.interface.process_control_interface` provide the ability to enable and disable the hardware device via
```python
set_activity_state(self, active: bool) -> None
```
Using the new mixin, this functionality can be accessed by the switch toolchain of qudi without additional implementation efforts for each hardware module.

Addresses part of the discussion in PR #31

## How Has This Been Tested?
On Win 10 pro 21H2 using the qudi-iqo-modules `default.cfg`

## Screenshots:
image

## Types of changes
- [ ] Bug fix
- [x] New feature

## Checklist:

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -9,6 +9,9 @@ None
 
 ### New Features
 - support for Zaber (linear) motorized stages (in hardware/motor/zaber_motion)
+- Added mixin `qudi.interface.mixins.process_control_switch.ProcessControlSwitchMixin` to provide 
+optional default implementation satisfying the `SwitchInterface` for process control hardware 
+modules implementing any of the interfaces contained in `qudi.interface.process_control_interface`
 
 ### Other
 None
--- a/src/qudi/default.cfg
+++ b/src/qudi/default.cfg
@@ -292,7 +292,7 @@ hardware:
             extend_hardware_name: True
         connect:
             switch1: 'switch1_dummy'
-            switch2: 'switch2_dummy'
+            switch2: 'process_control_dummy'
 
 
     switch1_dummy:
@@ -304,16 +304,6 @@ hardware:
                 one: ['down', 'up']
                 two: ['down', 'up']
                 three: ['low', 'middle', 'high']
-
-    switch2_dummy:
-        module.Class: 'dummy.switch_dummy.SwitchDummy'
-        options:
-            name: 'Second'  # optional
-            remember_states: True  # optional
-            switches:
-                'An even longer name of the switch itself':
-                    - 'Very long name of a random state'
-                    - 'Another very long name of a random state'
 
     fast_counter_dummy:
         module.Class: 'dummy.fast_counter_dummy.FastCounterDummy'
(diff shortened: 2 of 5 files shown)


GRexBayer:
With the dummy, this appears to work so far. However with the rework of the ni ao module there are some issues. Somehow the ao_voltage_range function in the helpers returns a tuple of length 4 (as well as the corresponding ni function which it wraps?) but it is expected to be of length 2. The returned tuple is (-5.0, 5.0, -10.0, 10.0) which to me/us does not make sense?

Neverhorst:
> With the dummy, this appears to work so far. However with the rework of the ni ao module there are some issues. Somehow the ao_voltage_range function in the helpers returns a tuple of length 4 (as well as the corresponding ni function which it wraps?) but it is expected to be of length 2. The returned tuple is (-5.0, 5.0, -10.0, 10.0) which to me/us does not make sense?

Yeah, I did not see that, thank you. Looking into the `nidaqmx` documentation is not very helpful either... they just claim this function returns "pairs of output voltage ranges supported by this device". So great...
I will not have access to a NI card for at least the next 2 weeks, so if you could manage to fix it somehow, we could merge earlier.

LukePiewalker:
Was just checking again on a simulted Ni and an Ni card at another setup. Apparently, on those two I only got a list of length 2 with [-10, 10] as expected. The version of nidaqmx was the same on the setup where we tested first, but maybe we should again check at that setup @GRexBayer?

LukePiewalker:
Ok, I finally spotted the difference.
It is the corresponding NiCard. With the PCIe 6363 (with which @GRexBayer and I originally tested) I get [-5.0, 5.0, -10.0, 10.0] and with a PCIe 6323  [-10.0, 10.0] even when just using simulated NI Cards. Also the spec sheet of the different models lists these in that form. Accoring to the manual, the 6363 can have two different AO references (10V and 5V) which causes this.

However, so far I have not found the command to get the current configured reference on the 6363, but I'll keep my eyes open.

GRexBayer:
I tested the module again with the same setup. Now, the error with the NiDaq does not occur when loading QuDi. However, I have issues connecting the switch interfuse with the analogue output module. When I attempt it, I get the error that the analogue output module does not implement the switch interface. With the pid controller dummy it works fine. I suppose I would need to implement the analogue output of the nidaq through a software pid to check the functionality?

GRexBayer:
> I tested the module again with the same setup. Now, the error with the NiDaq does not occur when loading QuDi. However, I have issues connecting the switch interfuse with the analogue output module. When I attempt it, I get the error that the analogue output module does not implement the switch interface. With the pid controller dummy it works fine. I suppose I would need to implement the analogue output of the nidaq through a software pid to check the functionality?

NVM, found the issue. I had to add the mixin interface to the hardware module, could fix it!

GRexBayer:
After adding the Mixin import to the analog output module, the switch interface works fine as intended. Tested by manually changing the output voltage of the NI card. Changing the voltage only works if the channel is set to active. Switching between active and inactive works fine.

GRexBayer (approved):
The Switch module was tested with a dummy as well as with a NI card using the analog output interface and works as intended.

PR #42, opened 2022-10-17
Save parameters of generated waveform

timoML:
With this change, the parameters of a generated waveform (eg. for generate_rabi: tau_start, tau_step, num_of_points) are saved with the measurement data. This makes it a lot easier to keep track of saved experimental results.

## Description
Exemplary saved file of a pODMR experiment:
image

Current catch (even prior to this change):
All meta-info saved from the pulsed module are the status variables at the time of hitting the save button. This doesn't necesarily correspond to the measured signal, as it is possible to change settings of the module after a measurement was finished.

## Types of changes
- [ ] Bug fix
- [x] New feature

## Checklist:


timoML:
@Neverhorst I made the generation_method_parameters their own entity, separate from ensemble.sampling_information and measurement_information as discussed. This requires to make the generation_method_parameters a property of pulsed_measurement_logic, which feels a little unelegant to me.

Neverhorst (approved):
Code looks good. Thank you.

PR #48, opened 2022-11-28
Ni interfuse data polling

LukePiewalker:
Rework of the data polling of the ni_scanning_interfuse.

## Description

The data poll now runs in a separate loop and the data poll is no longer triggered from the logic, as it should be. 

## Motivation and Context

Before, a data poll was only executed upon request from the logic even though the hardware might have had more data ready. This has been fixed and the poll now requests a minimum number of "chunk_size" samples but also gathers more if they are already available.

## How Has This Been Tested?

Tested on Win10 21H2 and a PCIe-6323 Ni Card connected to Galvo Scanner for x,y and an open loop piezo for z.
Tested with various clock frequencies and scan resolutions in 1D and 2D scans.

## Types of changes

- [x] Bug fix
- [ ] New feature

## Checklist:

- [x] My code follows the code style of this project. (Hopefully :wink:)

### DIFF
--- a/src/qudi/hardware/interfuse/ni_scanning_probe_interfuse.py
+++ b/src/qudi/hardware/interfuse/ni_scanning_probe_interfuse.py
@@ -116,7 +116,6 @@ class NiScanningProbeInterfuse(ScanningProbeInterface):
         self._default_timer_interval_ms = -1
         self._interval_time_stamp = None
 
-        self.__read_pos = -1
         self.__scan_stopped = False
 
         self._thread_lock_cursor = Mutex()
@@ -412,7 +411,6 @@ class NiScanningProbeInterfuse(ScanningProbeInterface):
             first_scan_position = {ax: pos[0] for ax, pos
                                    in zip(self.scan_settings['axes'], self.scan_settings['range'])}
             self._move_to_and_start_scan(first_scan_position)
-            self.__read_pos = 0
 
             return 0  # FIXME Bool indicators deprecated
 
--- a/src/qudi/hardware/interfuse/ni_scanning_probe_interfuse.py
+++ b/src/qudi/hardware/interfuse/ni_scanning_probe_interfuse.py
@@ -527,11 +527,14 @@ class NiScanningProbeInterfuse(ScanningProbeInterface):
 
     def _fetch_data_chunk(self):
         try:
-            self.log.debug(f'fetch chunk: {self._ni_finite_sampling_io().samples_in_buffer}, {self.is_scan_running}')
-            chunk_size = self._scan_data.scan_resolution[0] + self.__backwards_line_resolution
-            # chunk_size = 10  # FIXME Don't hard code chunk size? The above does not work with 1D scans. Why?
+            # self.log.debug(f'fetch chunk: {self._ni_finite_sampling_io().samples_in_buffer}, {self.is_scan_running}')
+            # chunk_size = self._scan_data.scan_resolution[0] + self.__backwards_line_resolution
+            chunk_size = 10  # TODO Hardcode or go line by line as commented out above?
+            # Request a minimum of chunk_size samples per loop
             try:
-                samples_dict = self._ni_finite_sampling_io().get_buffered_samples(chunk_size)
+                samples_dict = self._ni_finite_sampling_io().get_buffered_samples(chunk_size) \
+                    if self._ni_finite_sampling_io().samples_in_buffer < chunk_size\
+                    else self._ni_finite_sampling_io().get_buffered_samples()
             except ValueError:  # ValueError is raised, when more samples are requested then pending or still to get
                 # after HW stopped
                 samples_dict = self._ni_finite_sampling_io().get_buffered_samples()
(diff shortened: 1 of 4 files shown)


src/qudi/hardware/interfuse/ni_scanning_probe_interfuse.py
```diff
@@ -113,7 +116,6 @@ def __init__(self, config, **kwargs):
         self._default_timer_interval_ms = -1
         self._interval_time_stamp = None
 
-        self.__read_pos = -1
         self.__scan_stopped = False
```
timoML, line 119: I think we can drop __scan_stopped. It's only written, never checked in this module.

src/qudi/hardware/interfuse/ni_scanning_probe_interfuse.py
```diff
@@ -458,18 +466,9 @@ def get_scan_data(self):
...
-            if not self.is_scan_running or not self._ni_finite_sampling_io().is_running:
-                return self._scan_data
-            else:
-                # _stop_scan is called asynchronously. Thus .is_scan_running might be True, even if last data frame
-                # was already fetched. __scan_stopped is set by the polling thread, guaranteed to signal in time.
-                if not self.__scan_stopped:
-                    self._fetch_data_line()
-                return self._scan_data
+            with self._thread_lock_data:
+                return self._scan_data.copy()
```
timoML, line 471: Is it safe to call .copy()? Eg. what happens if no scan was configured before, _scan_data could be == None.

src/qudi/hardware/interfuse/ni_scanning_probe_interfuse.py
```diff
@@ -869,9 +858,53 @@ def __stop_ao_write_timer(self):
...
+            if self.number_of_scan_lines > 1:
+                reshaped_arr = self._raw[key].reshape(self.number_of_scan_lines,
+                                                      self.forward_line_resolution + self.backwards_line_resolution)
+                reshaped_2d_dict[key] = reshaped_arr[:, :self.forward_line_resolution].T
+            elif self.number_of_scan_lines == 1:
+                reshaped_2d_dict[key] = self._raw[key][:self.forward_line_resolution]
+        return reshaped_2d_dict
+
+    def backwards_data(self):
+        reshaped_2d_dict = dict.fromkeys(self._raw)
```
timoML, line 895: Heavy code duplication. Is backward_data even different from forwards data?

src/qudi/hardware/interfuse/ni_scanning_probe_interfuse.py
```diff
@@ -520,50 +519,40 @@ def _check_scan_end_reached(self):
         if self.__scan_stopped:
             return True
 
-        if self._scan_data.scan_dimension == 1:
+        if len(self.raw_data_container) == self._ni_finite_sampling_io().frame_size:
```
timoML, line 522: If we stick with the RawData structure in it's own class: Maybe a .is_full() property could be neat...

LukePiewalker: With feedback from @Neverhorst, probably the "RawData" structure will be dropped. But we can also discuss again today.

src/qudi/hardware/interfuse/ni_scanning_probe_interfuse.py
LukePiewalker: It indeed is not. So far the exception got swalloed by `self.log.exception("")`. But a simple `assert` could solve that and inform the user, that he should not call this without configuring a scan. `self._scan_data` is only None upon startup of the module.

src/qudi/hardware/interfuse/ni_scanning_probe_interfuse.py
LukePiewalker: Yep, you're right. Might use it though, when I will rework the "RawData" thingy again.

LukePiewalker: It was actually is checked. Yet "wrapped" under _check_scan_end_reached. But maybe you ment getting rid of that "wrapped" structure, which I did now.

timoML: In line 520 and 523, it's the only time that __scan_stopped is set True. Thus, all other code could simply call _check_scan_end_reached() to test this condition. I see no need to keep the class variable there then.

src/qudi/hardware/interfuse/ni_scanning_probe_interfuse.py
LukePiewalker: Did introduce the requested property in 97451e42. Even though the RawDataContainer will not last long after this PR.

src/qudi/hardware/interfuse/ni_scanning_probe_interfuse.py
LukePiewalker: I did completely remove _check_scan_end_reached and __scan_stopped so far and used a property of the RawDataContainer. Since, this will be gone at some point I did include the _check_scan_end_reached again in 1c522be yet using the property .is_full() so __scan_stopped is gone as requested.

timoML (approved):
Good for now.

PR #49, opened 2022-11-29
Fiy null bytes in Keysight module

kay-jahnke:
The keysight sometimes returns a null byte at the end of the string which crashes the manager GUI, so we strip it proactively.

Null bytes are a big NONO in python and should be avoided at all costs.

## Description

The origin of the null byte is the keysight hardware module, which we do not have access to, so we need to implement the fix null byte post call.

## Motivation and Context

Avoid a crash of qudi by catching unwanted null bytes from the hardware.

## How Has This Been Tested?

Testes on Hardware with current qudi core in anaconda on python 3.8.12.

## Types of changes

- [x] Bug fix
- [ ] New feature

## Checklist:

### DIFF
--- a/src/qudi/hardware/awg/keysight_M3202A.py
+++ b/src/qudi/hardware/awg/keysight_M3202A.py
@@ -152,10 +152,10 @@ class M3202A(PulserInterface):
             self.awg.close()
             raise Exception('AWG Error: {0} {1}'.format(aouID, ksd1.SD_Error.getErrorMessage(aouID)))
 
-        self.ser = self.awg.getSerialNumber()
-        self.model = self.awg.getProductName()
-        self.fwver = self.awg.getFirmwareVersion()
-        self.hwver = self.awg.getHardwareVersion()
+        self.ser = self.awg.getSerialNumber().rstrip('\x00')
+        self.model = self.awg.getProductName().rstrip('\x00')
+        self.fwver = self.awg.getFirmwareVersion().rstrip('\x00')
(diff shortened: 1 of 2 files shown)


src/qudi/hardware/awg/keysight_M3202A.py
```diff
@@ -152,10 +152,10 @@ def on_activate(self):
             self.awg.close()
             raise Exception('AWG Error: {0} {1}'.format(aouID, ksd1.SD_Error.getErrorMessage(aouID)))
```
Neverhorst, line 153: Are all placed handled where string buffers are used with the keysight DLL? Seems to me like `getErrorMessage` could suffer from the same issue. Basically all DLL calls that return a string should be checked as far as I understand.

kay-jahnke: Interestingly this is not the case: This error only appears because empty string buffers are created by the keysight and filled by `\0`.
The function `getErrorMessage` works with `c_char_p`, so no NULL filled buffers appear. 
All the other hardware calls in the hardware wrapper either return Int-like Status bytes or even proper ints, so this does not appear. A quick test on Hardware, with uploading sequences and playing them did not show any more NULL bytes.

Neverhorst: The inconsistency in Keysight engineering processes never ceases to amaze me. Ok, thanks. Looks good then.

Neverhorst (approved):

PR #53, opened 2022-12-20
Refactor scalar constraint

Neverhorst:
## Description
Got rid of any import of deprecated module `qudi.core.interface`.
Only the pulsed and time series toolchain were impacted by this change.

## Motivation and Context
`qudi.core.interface.ScalarConstraint` has been deprecated and was replaced by the quite different `qudi.util.constraints.ScalarConstraint`.
The new object has properties in place to provide the same interface as the old one.

## How Has This Been Tested?
Only on dummy. Testing is needed with real hardware for pulse generators (not critical due to minimal changes) and NI instreamer (critical due to interface change).

## Types of changes
- [ ] Bug fix
- [ ] New feature

## Checklist:

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -2,9 +2,17 @@
 
 ## Pre-Release
 ### Breaking Changes
+None
+
 ### Bugfixes
+None
+
 ### New Features
+None
+
 ### Other
+- Bumped `qudi-core` package minimum version requirement to v1.3.0
+- Got rid of deprecated `qudi.core.interface` module usage
 
 ## Version 0.4.0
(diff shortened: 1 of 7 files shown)


qku:
Hello, 
I tested this on my setup with an NI card. I can read off my APD count rate as usual with the time series gui, time series reader logic and ni_instreamer. The count rate changes as expected when I move around the cursor in the scanner gui. So I think the changes here should be fine! Let me know if you need additional testing.

Neverhorst:
Thank you very much @qku for testing. Now if someone would like to approve this PR, I can go on and merge it.

PR #69, opened 2023-01-29
Transition to PySide6

qku:
## Description

I played around with qudi-core and the dummies in qudi-iqo-modules trying to make them compatible with PySide6/Qt6.

Most changes are either
- replacements of `from PySide2 import ...` with `from PySide6 import ...`
- replacements of `QtWidgets.QAction` with `QtGui.QAction`
- changes to constants: something like `self.DockWidgetFloatable` had to be changed to `QtWidgets.QDockWidget.DockWidgetFloatable` to get it working

The more complicated changes are in 63a0a4f affecting the POI manager and the pulsed toolchain. Some signals seem to have changed in Qt6. Since I don't know the modules very well, I am not sure if the changes I made don't break anything. As an example, I found that I cannot delete POIs in the dummy anymore.

## Motivation and Context

I am sure that there are many good reasons to eventually transition to PySide6. Personally, I was motivated to play around with this since I am mostly developing on a laptop with Apple Silicon processor which [does not support PySide2](https://wiki.qt.io/Qt_for_Python).

## How Has This Been Tested?

Not very thoroughly: I loaded all dummy modules in the default config and pressed a few buttons at random.

## Types of changes

- [ ] Bug fix
- [X] New feature

## Checklist:

- [X] I have tested my changes using 'Load all modules' on the default dummy configuration.


qku:
See also the corresponding PR to qudi-core: https://github.com/Ulm-IQO/qudi-core/pull/45

TobiasSpohn:
Merged with #228

PR #70, opened 2023-02-01
Plugins to pulsed toolchain

timoML:
## Description
This PR introduces small changes to the pulsed toolchain that allow loaded predefined_methods more flexibility.
- Custom predefined methods inherid from `PredefinedGeneratorPlugin`, additionally to the existing `PredefinedGeneratorBase`.
- This enable to:
  1. Manipulate and even add their own `generation_parameters` (see screenshot)
  2. Invoke code after the `PulseObjectGenerator` is initialized. This allows for numerous possibilies, eg. to influence the helper methods in pulse_objects.py (like  _get_mw_element() ) that are used from all predefined methods.
- I also added the possibily to have enum like `generation_parameters` correctly handled by the gui.

## Screenshots:
Custom generation_parameters introduced by a predefined method plugin file.
image

## Types of changes

- [ ] Bug fix
- [x] New feature

## Checklist:


src/qudi/logic/pulsed/pulse_objects.py
```diff
@@ -1641,5 +1642,34 @@ def is_generator_class(obj):
...
+    PredefinedGeneratorPlugin is a PredefinedGenerator with addtional powers.
+    - It may manipulse the sequence generation parameters from it's code
+    - It can run code after the PulseObjectGenerator in order to manipulate all loaded predefined methods.
+    """
+    def activate_plugin(self):
+        # allow plugins to invoke code after the PulseObjectGenerator is fully initialized
+        pass
+
+    @property
+    def generation_parameters(self):
```
Neverhorst, line 1663: Maybe one could add some kind of interface method to `PredefinedGeneratorBase` to get/insert parameters from/into the `SequenceGeneratorLogic`.

Neverhorst (changes requested):
I think we should provide some additional interface to `PredefinedGeneratorBase`. Something like:

- `get_generation_parameters()` (returns a _copy_ of the underlying dict)
- `insert_generation_parameter(key, value)` (more transparent about what is going on)

If I understand this correctly, the property `generation_parameters` does not even need to be included in the `PredefinedGeneratorPlugin` class in this case because the new functions are inherited from the base class.

PR #71, opened 2023-02-27
On awg waveform generation

Geegee:
This is a mock PR to explore using remote pulsers. In this case the AWG70k Model B is implemented.
QuDi is run on the AWG (quiet specific case I guess).
## Description

In the case of remote access the FTP connection is not needed as the waveforms are generated on file.

## Motivation and Context

In this specific case we could not get the FTP running and thought we give it a try alongside 
the question if we can somehow speed up waveform generation.

## How Has This Been Tested?

Tested with AWG70k Model B. We also tried it then again with AWG70k Model A with FTP to see 
if everything is still running.

## Types of changes

- [ ] Bug fix
- [x] New feature

## Checklist:


TobiasSpohn:
Did not yield a speed improvement.

Issue #59, opened 2023-01-11
Wavemeter Instreamer Toolchain


ksenkalla:
### Feature Description

Rewrite the  wavemeter toolchain in such a way that the wavemeter can use the data instream interface.

### Related Problem

Porting from old core did not help as the "slow counter" doesn't exist anymore 

### Considered Alternatives

Alternative would be using the laser scanning directly using voltage to scan the laser. But this doesn't help to monitor the wavemeter

qku:
Hello, I tested the hardware file after moving to analog channels and it seems to work fine with the wavemeter I am using!

I saw that you increased the hardcoded number of digits behind the decimal in the time series gui. I guess it would be useful to have this as a config option or even accessible in the gui itself at some point.

ksenkalla:
> Hello, I tested the hardware file after moving to analog channels and it seems to work fine with the wavemeter I am using!
> 
> I saw that you increased the hardcoded number of digits behind the decimal in the time series gui. I guess it would be useful to have this as a config option or even accessible in the gui itself at some point.

Thats a good point, I will add this to the things we want to do.

> Regarding the plan from here: are we planning to stick to the hardware file implementing the instreamer interface? I understand that you have some plans for a custom logic and gui, but the hardware file will stay like this (with further improvements), right?

Yes exactly: For the future the hardware file can be adjusted so that all functions of the wavemeter can be covered. So on top of the already implemented methods further functions can be implemented (as e.g. setting the acquisition timing etc.)
We were setting it up like this with the notion that we try to standardize the modules, in sense of that any other wavemter can be used with the same toolchain. And we found that every wavemeter at least can read continuously some wavelength, which is basically an instream of wavelengths.

qku:
Hi @ksenkalla , I just played around with the latest state of your code. Unfortunately I cannot get the histogram GUI to work. What has to be running for it to work properly? How is it connected to the time series GUI/logic? At this point, I am pretty clueless about the idea behind this wavemeter histogram system.

qku:
Independently of the histogram gui, I think it would be cool to improve the wavemeter hardware module implementing the time series interface, e.g. by adding support for multiple wavemeter channels and units. I would be happy to work on it myself and contribute via a PR. A robust base can facilitate various wavemeter use cases, which may be rather diverse from one setup/experiment to another. Maybe it is even worth to consider merging a working hardware file into main already even if the histogram functionality is not finished yet.

qku:
[...]

PR #72, opened 2023-05-03
Wavemeter development: multi-channel, units, exposure time

qku:
## Description
This PR extends the functionality of the wavemeter as instreamer hardware module along with further improvements. A config option to select multiple channels on the wavemeter, the desired output unit and exposure time are added. The exposure time is set once during module activation.

Reading multiple channels of the wavemeter at once required some design considerations to be made. It is possible that only some channels have a new value during a `read_data_into_buffer` call. For this case, I decided to keep the last reading of each channel to be available as a kind of placeholder until a new value was sent from the hardware.

I realized that interpolation happens only if the sample rate is set to a low value and put in a warning for this case. I think it would be ideal to have only true timestamps and wavemeter readings in the data stream. Increasing the sampling rate should usually not be an issue.

In addition, I removed a few lines which I believe to be obsolete to make the module more readable.

## Motivation and Context

This is related to #59 where the wavemeter toolchain is discussed.

## How Has This Been Tested?

On a HighFinesse WS7 wavemeter with a 4-channel switch and two lasers attached, using the time series gui and logic.

Please let me know if this version of the module continues to work with the histogram gui. 

## Types of changes

- [X] Bug fix
- [X] New feature

## Checklist:


qku:
@ksenkalla @Nick-Grimm I had some free time for coding and decided to go ahead with extending the wavemeter hardware module with a couple features we discussed would be useful. I also made a few changes/additions/deletions which made sense to me, but might not to you, so do let me know what you think!

PR #79, opened 2023-06-23
Wavemeter data instream hardware file

qku:
## Description
This PR implements the HighFinesse wavemeter as an instreamer hardware module based on the work of @ksenkalla and others in the wavemeter development branch. The `TIMESTAMP` sample timing is used. A callback function receives all readings from the wavemeter. To enable readout of multiple channels from different modules (an thereby remote calls), all direct communication is done via a proxy module.

All files of the old wavemeter toolchain are removed. They were never functional with the new core anyways.

This PR also adds support for displaying nan values in the time series gui.

## Motivation and Context

This is the first step to resolve issue #59. The wavemeter hardware is now treated in a meaningful way as a device which provides a continuous stream of data points. It provides the foundation for more elaborate logic and gui modules. Multiple channels are fully supported, making sharing of a wavemeter between multiple labs possible.

## How Has This Been Tested?

This was tested with a WS7 wavemeter connected to two lasers through a multi-channel switch, with the time series reader logic and gui. To simulate readout from multiple (remote) machines, it was also testes with multiple hardware, logic and gui instances accessing different channels on the same wavemeter.

## Types of changes

- [ ] Bug fix
- [X] New feature

## Checklist:

- [X] I have tested my changes using 'Load all modules' on the default dummy configuration.
- [X] All changed Jupyter notebooks have been stripped of their output cells.


qku:
Have you had a chance to test the module yet @ksenkalla @Nick-Grimm ?

Nick-Grimm:
Hi @qku, In the meantime I have tested the module and quickly adapted the logic to retrieve the old functionality of laser scanning. However I tested so far only with a single channel, as our currently used wavemeter only supports one. I will try with multiple channels next week. Nonetheless, I have noticed a few things:

- First I tried to display the wavemeter data by the time series gui. Similar to you since only one channel is active the gui updates very slow. However, I think this issue is related to the time series logic and the way the data is read from the hardware. Thus for a proper logic this should not cause any problems.
- Further, as the time seris logic only uses method read_available_data_into_buffer is tried to use the other. For this I noticed that a call of read_data is malfunctioning as the parameter sampels_per_channel is missing & for me the 2x mutex causes additional problems.
- read_single_point does not return the most recent sample but the last entry of the data_buffer (thus 0.0) 
- The exposure time gets set during activation determined by the conifg. However, for me the window where the data aquisition is neither under- or overexposed is only up to a few ms. Meaning that we mostly work in the automatic mode, which changes the exposure time directly. I see that a constant "rate" is desirable, but Im not sure about this as a config option?
- And at last, as we have a option in the laser scanning gui to stop and resume the measurement I would suggest that stopping the wavemeter stream will not automatically reset the wm_start_time unless the users requests to. 

I think these are all minor issues and overall the module looks pretty neat. Let me know what you think.

Nick-Grimm:
Hi @qku!
> These two methods should work as expected now, thank you for thoroughly testing them!

Yes, the method read_data works as expected now. However, there is still one minor bug for the method read_single_point as it returns for one channel all available data (wavelength or frequency) and not just the single, most recent snapshot. 
> Can you maybe find a way to keep track of the timing you need in your logic alone?

Regarding the issue with the timings, I think there might be a workarround for this. However, I think for further application it might be convenient that stopping the stream rather pauses it, such that timings are conserved and can be resetted from the logic by stop_stream + setting wm_start_time to None. I am not sure if this conflicts the interface or makes using multiple channels more difficult. Also, as this is rather a matter of taste we can leave it as it is, but let me know what you think.  
Best regards!

qku:
Thank you for testing again @Nick-Grimm! read_single_point should work again now. I had simplified something and accidentally broke it that way.

qku:
Regarding the timings: I believe it makes sense from the general instreamer perspective that they should refer to the start of the stream. Otherwise somebody would have to keep track of the time while the stream isn't actually running. For the wavemeter this is the case. But even there, it is possible to stop acquisition itself in the wavemeter manufacturer GUI for example which resets the raw timestamps from the wavemeter. TLDR it would be difficult to rely on those timestamps as I see it.

qku:
@Nick-Grimm in case you are not aware: I plan to extend the hardware module further to enable simultaneous usage of a single wavemeter from multiple experiments. This requires the capability to have multiple instream instances run simultaneously which in turn means that I have to move out all callback related code to its own module.

I will let you know once that is complete. In the meantime, you could already ask around if someone could test the final module with a multi-channel wavemeter @ksenkalla :)

qku:
I was not able to read out the wavemeter remotely just yet, though. Since I get the same error also with the data instream dummy, I believe this not to be related to my code and therefore opened issue #84.

Nick-Grimm:
> @Nick-Grimm @ksenkalla I believe I am done with moving all communication with the wavemeter to a separate proxy module. This should enable readout from separate (remote) machines. I tested it by running two time series GUIs, logics, and hardware modules each accessing a different channel on the wavemeter.
> 
> As a side effect, the bug with the slow gui for a single channels seems to be resolved.
> 
> I would appreciate additional testing of the current state on your hardware!

Hi @qku,
Meanwhile I have tested the latest changes. I proceeded in the following way:

1. One channel high finesse wavemeter WS7/60 w/o and w/ remote readout.
2. (Buggy) multi-channel wavemeter WSU30 (two active switches) w/o and w/ remote readout & remote readout with two client PCs.
---
1. This case works well, except that an error message pops up: `Wavemeter error while activating channel 1: ResultError.NotAvailable` when activating the channel as this wavemeter does not support a switching mode. However, this did not restrained the functionality.
2. This wavemeter streams data very slow, but for the purpose of testing it should suffice. The cases of w/o and w/ remote readout went well. 
For the case that two client pc access the same remote module, I have in the channel config of the time series gui both channels active, bot only one channel visible, respectively. Then, simultaneous streaming went well, but as stopping the stream on the side of one client completly stops the acquistion of the high finesse application, the other time series gui stops as well with an error message that the device has stopped running. I wonder how your procedure was here, as this hinders shared acces on one wavemeter hardware. 

I hope this is useful for you.
And as you tested remote connections now, I would be interested in your procedure with the interplay on remote and client side and if you experienced problems with that? :)
Futher, I believe issue #84 as you added from qudi.util.network the netobtain can be closed once there exists proper documentation?

qku:
I am happy to hear that the modules are working well overall also on your side. Thank you for testing @Nick-Grimm !

Since I don’t have a wavemeter without switch, I will not be able to reproduce the error you got in scenario 1. I believe to know where it occurs though and will implement some changes later this week to avoid this error.

Acquisition of the wavemeter should only stop once the last instreamer module stops streaming. Can you try deactivating the channels you don’t use in the respective time series GUIs?

I agree that issue #84 can be closed with this PR. I appreciate the documentation you are doing on remote modules!

Nick-Grimm:
Hi @qku,
@LukePiewalker and I now tested with a properly working wavemeter. For this wavemeter, we first observed an issue that our wlm-version did not support the python high_finesse_wrapper for the wlmData.dll. It would be convenient if for this case can be checked during loading of the dll.
With another wavemeter supporting multi-channels we then could successfully test two instances of the wavemeter to be read out remotely and simultaneously by the timeseries gui. Indeed, only the last instreamer module stops the streaming. 👍

qku:
Thank you for the additional testing @Nick-Grimm @LukePiewalker !

> For this wavemeter, we first observed an issue that our wlm-version did not support the python high_finesse_wrapper for the wlmData.dll. It would be convenient if for this case can be checked during loading of the dll.

What error do you get exactly? We could then add an additional exception clause and warn the user accordingly. And out of curiosity, how did you resolve this? Did you update your DLL?

qku:
I have also tried to improve compatibility with wavemeters without switch by checking once if a switch is present in the beginning. Maybe you could give these changes another quick test @Nick-Grimm to see if you now get no unnecessary error messages anymore. I would do it myself if I only had access to a suitable device. Thank you!

LukePiewalker:
Hi @qku, 
> What error do you get exactly? We could then add an additional exception clause and warn the user accordingly. And out of curiosity, how did you resolve this? Did you update your DLL?

The solution was actually quite simple: We went for our second wavemeter with multiple channels/switch 😄.
There, the software is at a newer state.

> What error do you get exactly?

The error we got was that some function wrapped was not existing.
`AttributeError: function 'GetExposureNumEx' not found`

To resolve one could check the version number with `GetWLMVersion` to be higher then the lowest one we have tested to work. With Version 1673, we got this issue, yet its a very old version I suppose and I/we plan on getting an update here.
Out of curiosity what version do you use? Cause for me the manual does not even mention the function `GetExposureNumEx`? 

Regarding the other part you mention in your PR about the name change in the interface, I would highly advise to do this in a separate PR, even though its such a tiny change. However interface changes almost always bring along breaking changes and one would need to adjust all existing hardware to this (and some of it we might not even know of, in case it resides in some other fork).
I've seen you did account for the `InStreamDummy`, yet e.g. the `NiInstreamer` would still be missing.

The `use_own_data_buffer` thing seems a little hacky to me. But that's more of a gut feeling then a founded opinion. Its needed cause in case of a remote connection, it will not work to read the data into the local data_buffer passed as a reference to the streamer correct? Maybe @Neverhorst has some idea if there is a better way? Or if he approves I shut up about this one 😄.

LukePiewalker:
Also we encountered some weird behavior for the remote modules, e.g deactivation caused some bugs, and a deactivated module on the host side could still be accessed and operated from the client, yet I will have a deeper look, if this is maybe something core related.

Nick-Grimm:
> I have also tried to improve compatibility with wavemeters without switch by checking once if a switch is present in the beginning. Maybe you could give these changes another quick test @Nick-Grimm to see if you now get no unnecessary error messages anymore. I would do it myself if I only had access to a suitable device. Thank you!

Hi @qku, I have tested again with the single channel device and don't get the unneccessary error message any more. 👍

qku:
Thank you for the feedback @LukePiewalker !
> To resolve one could check the version number with `GetWLMVersion` to be higher then the lowest one we have tested to work. With Version 1673, we got this issue, yet its a very old version I suppose and I/we plan on getting an update here. Out of curiosity what version do you use? Cause for me the manual does not even mention the function `GetExposureNumEx`?

I implemented a version check now and separated loading the DLL itself from setting up its arguments to enable finer exception checking. The version we are using is 6533. What is the oldest version you got to work @Nick-Grimm @timoML ? Interestingly, the function `GetExposureNumEx` is also not documented in my manual but is available from the DLL.

qku:
> Regarding the other part you mention in your PR about the name change in the interface, I would highly advise to do this in a separate PR, even though its such a tiny change. However interface changes almost always bring along breaking changes and one would need to adjust all existing hardware to this (and some of it we might not even know of, in case it resides in some other fork). I've seen you did account for the `InStreamDummy`, yet e.g. the `NiInstreamer` would still be missing.

You are right. I undid the changes with my last commit and will consider to open another PR in the future.

LukePiewalker:
> I implemented a version check now and separated loading the DLL itself from setting up its arguments to enable finer exception checking. The version we are using is 6533. What is the oldest version you got to work @Nick-Grimm @timoML ? Interestingly, the function `GetExposureNumEx` is also not documented in my manual but is available from the DLL.

The revision number (I assume that's also what your referring to with the "6533"; I was originally a bit confused by revision number and version number) in which the `AttributeError` did not occur was **6491**.
Under revision number 6425, which is the other wavemeter we had at hand, we got the Error.

From a not too deep glance to the code, I suppose the module should also be able to run in older versions, since there are actually not too many dll functions in real use? Maybe I'll check if it would still tun with the old version with a "minimal load" of the functions in use, hopefully soon.
However I suppose wrapping all functions from the dll, like you did, is the cleaner and proper way to go. However, I have neither an informed opinion nor knowledge about that.

LukePiewalker:
> As far as I understand, passing a shared buffer which is filled with values does not work over rpyc. I might be wrong though as I am far away from being an expert. I now implemented an automatic check to see if the streamer hardware is a remote module. This allows us to at least get rid of the extra config option. Does this resolve your concerns @LukePiewalker ?

I'm also far from being an expert here, which is why I tried to get @Neverhorst  involved, since this is something general affecting all `InStreamer` Hardware which is running remotely. Or more general, if there is  any clever way of how one could handle this "passing a buffer" thingy to a remote module.

LukePiewalker:
Will also try to give the whole hardware another test, maybe within this week. Cannot promise however.

Neverhorst:
Well, communicating with remote modules is inter-process communication after all. The most efficient way to fetch data from the NI card is passing a buffer raw pointer to the driver but this will not work if the buffer is owned by another process.
The only way to work around such issues (and many other), is smarter code that handles the peculiarities of the remote connections for each case (see e.g. the use of `netobtain` all over the place).

So yes, in my opinion the conditional buffer handling using the `self._streamer_is_remote` flag seems justified.
You will notice that the remote module case will severely limit the maximum stream data rate, but that is just something we have to live with probably. Another approach could be to have the logic also run on the remote machine. There is far less data rate needed for `GUI <--> logic` communication compared to `logic <--> hardware`.

Keep also in mind that using remote references (rpyc proxies) to `numpy` arrays without copying them (netobtain) is fine in many cases (not here), depending on how you access them. This is also usually faster than using `netobtain`.

qku:
Thank you @LukePiewalker for looking up the revision number, I updated the code accordingly. What we could do is put the `argtypes` and `restype` setters of each function into their own try-except clause. That way the wrapper could still be used if not all functions are supported. I have to say that I am too lazy to do that though, considering the option to upgrade the wavemeter software.

LukePiewalker:
> Is there anything left to do here once you tested the hardware another time? I can't think of anything from my side.

From my side, I don't think so.
I was able to quickly test (partially) but was running into some confusion due to a tiny bug in the docstring of the `read_data` function. That's also in the interface. That kind of side tracked me a bit.
It states that with `data_buffer.reshape([<channel_count>, samples_per_channel])` the data could be unraveled, however I think it should be the other way around from what I afterwards checked with the Instreamer Dummy.
In `read_data_into_buffer` its the other way around.

So I only tested the wavemeter on a local machine (no remote) with two channels in use and could not really find any bugs.
I only noticed that the timing was a bit off. 60s real time were like ~50s on the x-axis in the timeseries gui.
But I could not really track down wether this is due to the nature of `SampleTiming.TIMESTAMP` or due to some rounding error in the `sample_rate` or similar. I also don't know wether this should be of any concern? @Nick-Grimm What do you think on this?

@qku Thanks a lot for your work and contribution by the way!!!

LukePiewalker:
> Thank you @LukePiewalker for looking up the revision number, I updated the code accordingly. What we could do is put the `argtypes` and `restype` setters of each function into their own try-except clause. That way the wrapper could still be used if not all functions are supported. I have to say that I am too lazy to do that though, considering the option to upgrade the wavemeter software.

Also thougth about the `try`/`except` part, yet also had been to lazy yet :wink:. I guess that part is fine from my side.

qku:
> From my side, I don't think so. I was able to quickly test (partially) but was running into some confusion due to a tiny bug in the docstring of the `read_data` function. That's also in the interface. That kind of side tracked me a bit. It states that with `data_buffer.reshape([<channel_count>, samples_per_channel])` the data could be unraveled, however I think it should be the other way around from what I afterwards checked with the Instreamer Dummy. In `read_data_into_buffer` its the other way around.

You are correct, the docstring is wrong. I just checked the `reshape` method with an instream dummy and `data_buffer.reshape([<samples_per_channel>, <channel_count>])` correctly produces an array with one column of samples per channel. I fixed the docstring with my last commit in the interface, the dummy and the wavemeter hardware.

Nick-Grimm:
> So I only tested the wavemeter on a local machine (no remote) with two channels in use and could not really find any bugs. I only noticed that the timing was a bit off. 60s real time were like ~50s on the x-axis in the timeseries gui. But I could not really track down wether this is due to the nature of `SampleTiming.TIMESTAMP` or due to some rounding error in the `sample_rate` or similar. I also don't know wether this should be of any concern? @Nick-Grimm What do you think on this?

I reproduced this timing mismatch also on our single channel wavemter. However, as I quickly went through the code I think this issue is related to the `time_series_reader_logic`. My gut feeling tells me that here the number of samples `samples_to_read` which the logic requests the hardware to read is higher (due to `_samples_per_frame`) than the 'available samples' as these are countinously read out. Thus, as the 'read_data_into_buffer' is written in such a way that it waits until the available samples again statisfies the requested samples the timing mismatch can occur. 
However, I don't think this is of bigger concern as at least I think we are using the time_series logic/gui rather to quickly test the hardware module. In a more specified logic (such as we are using it at the moment) we take more care about the timing.      

So, as far as I am aware there isn't anything left to test! 👍

LukePiewalker:
> However, I don't think this is of bigger concern as at least I think we are using the time_series logic/gui rather to quickly test the hardware module. In a more specified logic (such as we are using it at the moment) we take more care about the timing.

Since the Wavemeter is written as an Instreamer Type Hardware, I think it should behave as expected when read out with the time series toolchai, shouldn't it?
Unfortunately, I have no capabilities to test it this week.
I'm also not really grasping how the timing should be messed up in the end. Cause the hardware is supposed to deliver data and timestamps right? And the timestamp should encode the information of when the data is read which should be decoupled from how many samples are requested?
So either, there is something going wrong in the toolchain with Timestamp timing, or the timestamps are off a bit when retrieved from the hardware module?

Nick-Grimm:
> I'm also not really grasping how the timing should be messed up in the end. Cause the hardware is supposed to deliver data and timestamps right? And the timestamp should encode the information of when the data is read which should be decoupled from how many samples are requested? So either, there is something going wrong in the toolchain with Timestamp timing, or the timestamps are off a bit when retrieved from the hardware module?

Yes, I think you are right on this one and it is not as simple as I thought at the beginning. Having a closer look, I think the issue is related with the Timestamp toolchain and how the `_trace_times` array is initialised and treated when displaying it in the Gui. As we do not exactly acquire the data per seconds according to the data rate, we accumulate mismatch in the `_trace_times` array. Thus, meaning when in the Gui function `update_data` the `data_time` array is recalculated by substracting the continously changing shift time `data_time[0]` we wrongly calculate the timings that getting displayed.

When running the acquisition with a trace length of 10s shortly after the beginning the `_trace_times` may look like this:
 `In[5]: time_series_reader_logic.trace_data[0]
Out[5]: array([-7.888, -7.88 , -7.872, ...,  3.033,  3.043,  3.055])`,
while when running it for longer:
`In[6]: time_series_reader_logic.trace_data[0]
Out[6]: array([64.556, 64.567, 64.578, ..., 79.031, 79.042, 79.054])`    
Hence, here 79-64.5 = 14.5, the latest value does not get displayed in (at least) the live plot. I think for the recorded plot and data this is not an issue, but I haven't really checked yet.

qku:
I did a lot of testing to track down the source of incorrect timing in the time series gui. It seems like there are only issues if the sample rate estimate does not match the actual rate of incoming data from the wavemeter.

Once a channel is activated (deactivated) on the wavemeter switch during an ongoing stream (e.g. from another remote client), the rate of incoming wavelength measurements can drop (increase) dramatically. It is therefore very difficult to keep the sample rate estimate constantly accurate.

In my opinion, the only way to make this work nicely and robust is to make the time series toolchain independent of the sample rate estimate for timestamped data (issue #91). This should be possible, but will require some work which I consider to be beyond the scope of this PR.

LukePiewalker:
@qku, that sounds reasonable to me thus far.
However, I would like to point to the first concern you had today for raising exceptions, yet now for the warning case:
I think we should take care that one is not continuously spammed by warning messages in these cases, cause e.g. an underexposure is not fixed by itself, and hence the warning will appear quite often.

I think I would be also fine with all of the cases resulting in some exception and stopping the stream, if there is no clever way to handle these warnings in order to not spam the qudi log.

From my lab experience, we pretty much always have the wavemeter running in auto exposure, so under/overexposure is rare, if it happens at all.

From a quick glance at the `TimeSeriesReaderLogic` in case of an exception, the streamer will be stopped, as well as the data polling. Hence, I guess other then raising exceptions with meaningful error messages, there is nothing left to do on the hardware side I suppose?

qku:
> @qku, that sounds reasonable to me thus far. However, I would like to point to the first concern you had today for raising exceptions, yet now for the warning case: I think we should take care that one is not continuously spammed by warning messages in these cases, cause e.g. an underexposure is not fixed by itself, and hence the warning will appear quite often.

That is indeed important. A flooded log is really annoying and will make any kind of diagnosis difficult.

> I think I would be also fine with all of the cases resulting in some exception and stopping the stream, if there is no clever way to handle these warnings in order to not spam the qudi log.

This is certainly the easiest way to deal with these and should fit most use cases. It is also the safest, as stopping the stream cannot go unnoticed while a lot of under/overexposed samples in a stream can.

> From my lab experience, we pretty much always have the wavemeter running in auto exposure, so under/overexposure is rare, if it happens at all.

Good point. I have not used auto exposure a lot so far, but using it further stresses the need for enhancements in the time series toolchain as described in #91.

> From a quick glance at the `TimeSeriesReaderLogic` in case of an exception, the streamer will be stopped, as well as the data polling. Hence, I guess other than raising exceptions with meaningful error messages, there is nothing left to do on the hardware side I suppose?

I suggest to remove the conversion to nan from `high_finesse_proxy` and raise a custom exception in `high_finesse_wavemeter`. Since that will occur in between polls of the logic for new samples, it won't be caught by the logic though. I therefore suggest that the hardware stops its own stream upon detection of an error code / negative value.

qku:
I also noticed that I usually not raise exceptions in the hardware file, but log errors instead. @Neverhorst This does not follow our guidelines for hardware modules, correct? I think I will change those to proper exceptions like `ValueError` and `RuntimeError`.

Neverhorst:
It's fine for the hardware to stop independently if it is running in its own thread. However you would need to notify the logic of this error the next time it interacts with the hardware as if it was running (e.g. when polling data) in order to allow the logic module to tear down the measurement in an orderly fashion. So a hardware-internal error flag of some sort could be used to "cache the error" until the logic can receive it.
The result would most likely be that the logic is then trying to stop the hardware "again" during exception handling but this should be fine.

Regarding your last question @qku :
> I also noticed that I usually not raise exceptions in the hardware file, but log errors instead. @Neverhorst This does not follow our guidelines for hardware modules, correct? I think I will change those to proper exceptions like `ValueError` and `RuntimeError`.

Logging errors in qudi modules instead of raising exceptions can be a viable option, especially for threaded logic modules where slots are called through queued connections (Qt). In this case an exception is not propagated to the caller (signal emitter) and the only possibility to make the exception visible is `self.log.exception()` or `self.log.error()`. 
Public logic module methods must usually be able to handle exceptions when being called via queued connection (e.g. GUIs) AND direct calls (e.g. notebooks or console). This is an unresolved issue with qudi and can not be handled in a general way for now.  
For hardware modules this is usually no issue because their interface methods are accessed by direct calls from a logic thread (or a notebook/console) so raising exceptions can be considered general good practice for hardware modules.

qku:
As it turns out, it is difficult to have the wavemeter instreamer stop its own stream.

The instreamer hardware module receives new values via a call of its `process_new_wavelength` method by the hardware callback function. If `stop_stream` is called from there, and if the channel is the only one still active on the switch, the hardware module will notify the proxy to stop the callback procedure as it is not required anymore. Obviously this doesn't work - the callback procedure cannot delete itself. At least that is how I am diagnosing the issue I am facing right now where the deregistering of the callback function doesn't go through.

Personally, I would be _very_ happy if we could simply keep the error values in the data stream as they are and let the time series GUI and logic display them for now. Further processing is then up to the user/developer of more sophisticated logic. These error values usually just stand for over/underexposure and are not critical in the sense that the data stream doesn't stop. I therefore do not consider stopping the data stream upon encounter of one of those a must-have feature.

qku:
We could still improve on the current state though - what do you think about converting erorr codes to nan, but logging a warning only if the last sample did not contain an error code? That way we don't get strange negative values in our data stream but still have a change of finding out what exactly went wrong during the measurement.

Something similar is done for buffer overflow to avoid flushing the log.

qku:
Let me know what you think of this way to handle things @Neverhorst @LukePiewalker

Neverhorst:
> As it turns out, it is difficult to have the wavemeter instreamer stop its own stream.

Without having digged deeper into the code, I agree that it is probably extremely difficult (if not impossible) to stop the device from within the callback itself. However there are a few tricks one can employ to make the actual qudi thread stop it instead.
All ideas that currently come to my mind are based on a watchdog that runs in a separate thread:
- It might be possible for the callback to set a simple flag inside the hardware module that the watchdog can monitor and intervene to stop the hardware in the regular way (same way the logic does it)
- Let the watchdog monitor the data stream and stop it in case of a certain value (e.g. NaN)
- Have a second error code data stream analogous to the data stream that is invisible to the logic module but monitored by the watchdog
- Use OS features to set a flag that can be shared between threads/processes and monitored by the watchdog (could even be a file... but that is probably bad design)

qku:
Actually I now believe it is a bad idea to stop the stream when an error code is encountered.

Imagine a long running experiment with an unstable laser that might lead to a small percentage of failed wavemeter measurements. This can occur even with auto-exposure if the power fluctuations are very large. In general, it will not always be sensible to stop the entire measurement then as you may very well be able to get good data out by post-selecting the valid readings, don't you think?

In the implementation I propose, error codes are converted to NaN and the reason for error is logged. This should satisfy most of use cases in my opinion.

LukePiewalker:
As long as the error does not happen silently and the user is notified somehow that something is going wrong it is fine for me.

In the old core the stream was also continuing, yet the negative value was displayed and the user knew something is weird. So for me the proposed solution sounds fine.

LukePiewalker:
I gave it some test (without a remote module in this comment) and found some issues, however I do not know yet how to solve them properly or if they are of concern.

- In case one channel of the switch is underexposed the log is heavily spammed with the warning. For both underexposed (I tested with 2 channels configured) the warning is only occurring once.
- Interaction with the GUI of the wavemeter manufacturer can result in qudi freezing. Mostly encountered issues in this case, when using the `TimeSeriesGui`. Maybe some error happens silently and we are stuck in some loop? I can reliably reproduce the unwanted behavior, when deselecting the "Use" toggle for one of the wavemeter channels while the `TimeSeriesGui` is running and then trying to stop the stream via the `TimeSeriesGui`. Activating the "Use" button again brings the `TimeSeriesGui` back to life.

LukePiewalker:
> In general, it will be difficult to be robust against any changes in the manufacturer GUI. We can try to catch more codes in the callback as we are doing already for a click on "stop measurement". Disabling of a channel is difficult to catch though as only some internal exposure times change.

For this part, I leave it up to you to judge whether this is possible, since I do not have the resources right now to dive into the code and the manual. But does a deactivated channel still return some meaningful value? Or is the call just completely blocking and therefore causing the freeze?

> I believe that the only truly safe way of operation would be to disable the manufacturer GUI entirely.

Disabling the GUI I would personally try to avoid, since I do recall use cases where people did use the same laser for alignment purposes and needed to have quick glances on the current wavelength.
Using Qudi for that seems a bit overkill to me.
I guess people could be able to adapt to "Don't push buttons in the GUI" at some point.
However this is again a case of something failing silently, so I still have some mixed feelings about this behavior.

qku:
I implemented a watchdog 🐶 which checks for unwanted interaction with the manufacturer GUI in two ways:

- the callback function writes to an error flag if it is notified by the DLL (stop acquisition clicked, switch mode deactivated, ...)
- the watchdog compares itself if the channels activated on the hardware match the channels of connected instream modules

It then stops all streams of all instreamers. To enable proper teardown of the watchdog, I turned the proxy into a proper qudi module. A change of your config files is therefore required @LukePiewalker @Nick-Grimm.

This works with multiple instreamers, and also remote instreamers in 99% of all cases. I only rarely observed a freezing remote stream but could not track down the issue.

LukePiewalker:
Thanks for that  @qku

> This works with multiple instreamers, and also remote instreamers in 99% of all cases. I only rarely observed a freezing remote stream but could not track down the issue.

Is the stream freezing without interaction, or only after some specific interaction? And is it then blocking completely or just freezes for some time and unfreezes automatically?

qku:
> Is the stream freezing without interaction, or only after some specific interaction? And is it then blocking completely or just freezes for some time and unfreezes automatically?

I couldn't find a repeatable pattern to cause the freeze. The symptoms are clear though: the data in the time series gui on the client stops scrolling. A click on the stop button causes all buttons to grey out. Recovery is only possible though forcing qudi to quit. The server and the time series gui there continue to work fine though.

LukePiewalker:
> I couldn't find a repeatable pattern to cause the freeze. The symptoms are clear though: the data in the time series gui on the client stops scrolling. A click on the stop button causes all buttons to grey out. Recovery is only possible though forcing qudi to quit. The server and the time series gui there continue to work fine though.

I am just now running a test with a remote instreamer (NI) to see if it is some general remote issue. However the sample_rate is fixed, so it might not be 100% conclusive, if there is no freezing here.

While doing so, found a tiny inconsistency in the Instreamer and TimeSeries toolchain, causing me to post PR #103. Line 508 in [`TimeSeriesReaderLogic`](https://github.com/qku/qudi-iqo-modules/blob/wavemeter_impr_instream/src/qudi/logic/time_series_reader_logic.py#L508C106-L508C106) was causing the issue with a remote `NIXSeriesInStreamer`.

LukePiewalker:
I managed to freeze it with a remote `NIXSeriesInStreamer` as well after ~1h of Runtime.
 Somehow the "client disconnected from the module service" (qudi log info on the host PC) and I got the warning (in the qudi log of the client) "Reading data from streamer went wrong: [WinError 10054] Eine vorhandene Verbindung wurde vom Remotehost geschlossen". So this could be unrelated to the wavemeter.

So I assume there might be some underlying issue in the core which causes this. Could you check if you get the same error message on the host when it freezes for you @qku?
Any Idea where this could originate from @Neverhorst?

qku:
I will check if I get a "client disconnected" or any other log message when it freezes later this week.

Nick-Grimm:
I tested this PR with the single channel wavemeter locally as well as remotely. So far, I could not observe any "freezing" when stopping the stream (also >1h) or due to any other interaction. Further, implementation of the watchdog with the PLE scanning logic seems to work fine and without freezing.

qku:
I tested again today with a remote wavemeter. Even after four hours of running, I could not freeze anything.

I suspect that the freeze I initially observed was due to some problems in our lab network. I don't know of a better explanation. Therefore I consider this PR ready to merge.

Nick-Grimm (approved):
This PR has implemented the HighFinesse wavemeter as an instreamer module for qudi. It works with the time series reader toolchain and provides foundation for more elaborate logic and gui modules. Note that for timestamped data in the time series toolchain, estimation of the sample rate is still required (see open issue #91). This PR has been tested extensively on three different devices. Since the rarely observed freezing could not be tracked down to be an issue of the wavemeter, I am approving this PR ready for merge. 
A huge thanks to @qku for the contribution and the patience to consider all comments.

PR #80, opened 2023-06-26
Bugfix in sequencegeneratorlogic

Geegee:
We faced the issue that sometimes we could not reconstruct our experiment because 
the generation parameters were not saved. After some investigation it turned out 
that this happened in the case of granularity problems. 

Why is github showing ten commits although I just went into the main branch updated it and created a new branch
from there?

## Description
Small bugfix.

## Motivation and Context
Loss of measurement information

## How Has This Been Tested?
With an ipython notebook with several different generation parameters. 

## Types of changes

- [ x] Bug fix
- [ ] New feature


Neverhorst:
Please clean up your branch and re-submit this PR. Lots of unrelated "camera-stuff"...

PR #81, opened 2023-06-26
Bugfix in sequence_generator_logic.py.

Geegee:
We faced the issue that sometimes we could not reconstruct our experiment because 
the generation parameters were not saved. After some investigation it turned out 
that this happened in the case of granularity problems. 

## Description
Small bugfix.

## Motivation and Context
Loss of measurement information

## How Has This Been Tested?
With an ipython notebook with several different generation parameters. 

## Types of changes

- [ x] Bug fix
- [ ] New feature


timoML (approved):
Ok, I now see what's going on (see comment I added). Thanks for finding this nasty one!

PR #82, opened 2023-07-03
Fix Counts per second in finite sampling input

LukePiewalker:
Fix values retrieved from Ni Finite Sampling input to be "counts/second"

## Description

So far the Finite Sampling Input hardware module of the NI returns digital values which corresponded to "counts per clock cycle" and not "counts per second", as it is expected to be (and also as stated in the default config)

## Motivation and Context
From the old qudi core, the unit of the values gotten from digital channels had typically been counts/second. So far the unit was and still is ignored completely. Maybe one should not even allow to configure this and always return the value in c/s? 

## How Has This Been Tested?
Tested on a simulated NI PCIe-6363 with up to date qudi-core.

## Types of changes

- [x] Bug fix
- [ ] New feature

## Checklist:

### DIFF
--- a/src/qudi/hardware/ni_x_series/ni_x_series_finite_sampling_input.py
+++ b/src/qudi/hardware/ni_x_series/ni_x_series_finite_sampling_input.py
@@ -210,6 +210,7 @@ class NIXSeriesFiniteSamplingInput(FiniteSamplingInputInterface):
 
         # initialize default settings
         self._sample_rate = self._constraints.max_sample_rate
+        # TODO: Get real sample rate limits depending on specified channels (see NI FSIO), or include in "ni helper".
         self._frame_size = 0
 
         self.set_active_channels(digital_sources.union(analog_sources))
--- a/src/qudi/hardware/ni_x_series/ni_x_series_finite_sampling_input.py
+++ b/src/qudi/hardware/ni_x_series/ni_x_series_finite_sampling_input.py
@@ -424,7 +424,9 @@ class NIXSeriesFiniteSamplingInput(FiniteSamplingInputInterface):
                     timeout=self._rw_timeout)
                 if read_samples != number_of_samples:
                     return data
-                data[reader._task.name.split('_')[-1]] = data_buffer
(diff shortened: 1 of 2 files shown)


src/qudi/hardware/ni_x_series/ni_x_series_finite_sampling_input.py
```diff
@@ -424,7 +425,9 @@ def get_buffered_samples(self, number_of_samples=None):
                     timeout=self._rw_timeout)
                 if read_samples != number_of_samples:
                     return data
-                data[reader._task.name.split('_')[-1]] = data_buffer
+                data[reader._task.name.split('_')[-1]] = data_buffer * self._sample_rate
```
Neverhorst, line 428: To avoid another copy of the `data_buffer` array, it would be better to split this line into two statements, a multiplication (in-place array operation) and the assignment to the `data` dict, e.g.:
```python
data_buffer *= self._sample_rate
data[reader._task.name.split('_')[-1]] = data_buffer
```

The buffer handling is already pretty bad/taxing (written by myself). The NI in-streamer I recently wrote performs much better.

Neverhorst (changes requested):

NicolasStaudenmaier:
Tested on a running setup with NI USB-6343. CW ODMR now shows the correct c/s as expected from the instreamer. Can be approved from my side as soon as the requested change by @Neverhorst  is done.

src/qudi/hardware/ni_x_series/ni_x_series_finite_sampling_input.py
LukePiewalker: Thanks for this feedback, is implemented that way in https://github.com/Ulm-IQO/qudi-iqo-modules/commit/f1eb30f51840e572aadeb8bb49df7abbf00add44.

LukePiewalker:
@Neverhorst and @NicolasStaudenmaier,
I changed as requested from @Neverhorst. Is it ready to merge now?

Neverhorst (approved):

PR #96, opened 2023-09-06
Increase Keysight AWG sample rate precision 

NicolasStaudenmaier:
Keysight AWG allows a sample rate precision of 10mHz. So far the sample rate can only be set with 10MHz precision.

## Description

One line change to increase the number of digits where the sample rate is written to the AWG. 

## Motivation and Context

The AWG allows for a higher resolution of sample rates than is used in Qudi so far. People might want (including at least me) to set the sample rate with higher precision. 

## How Has This Been Tested?

Tested with Keysight AWG M8195A on my setup. 
Might require testing with model 8190A (@timoML?). 

## Types of changes

- [x] Bug fix
- [ ] New feature

## Checklist:


timoML (approved):
Looks good, thank you.

PR #97, opened 2023-09-08
scanning probe dataclasses and back scan functionality

### LINKED ISSUE #63: Implement "backscan" functionality with ni interfuse.
### Feature Description

One should be able to configure the amount of points which are scanned "on the way back". This is especially necessary if you have some sequence of scanning going on, like e.g. in the optimizer toolchain which has different scan resolutions and frequencies.

### Related Problem

So far, the amount of points scanned while moving "backwards" in a 2D scan is hardcoded. So in case of the optimizer with very small amount of pixels, the harcoded "__backwards_line_resolution" is too much (more pixels scanned on the way back) but one might want more for a regular scan.

### Considered Alternatives

Leaving it how it is right now is no option, since it has been implemented that way to reduce workload and have some somehow running solution quick. Also it was originally intended to be in the toolchain.

### Additional Context

_No response_

### Contact Details

_No response_

### LINKED ISSUE #64: Scanning probe interface still uses boolean indicators for failures etc.
### Feature Description

Rather then using int return values or boolean failure indicators, one should raise meaningfull exceptions if something goes wrong.

### Related Problem

Current implementation is a mess between, bool and int indicators which does not necessarily help while bug fixing.

### Considered Alternatives

This field is mandatory. But there is no real Alternative, then doing it proper.

### Additional Context

_No response_

### Contact Details

_No response_

### LINKED ISSUE #90: [New Feature] scanning probe toolchain: ScanSettings
### Feature Description

Use separate keyword arguments for axes, range, resolution and frequency in `configure_scan` in `scanning_probe_interface` instead of a dictionary. At the moment, the dictionary entries are not documented.

### Related Problem

_No response_

### Considered Alternatives

- passing a custom data class
- documenting the dictionary entries

### Additional Context

_No response_

### Contact Details

_No response_

qku:
### Interface
- add new `ScanSettings` dataclass to hold all settings of a scanning probe measurement (closes #90)
- decorate `ScannerChannel`, `ScannerAxis`, `ScanConstraints` and `ScanData` with [`dataclass`](https://docs.python.org/3/library/dataclasses.html#), replace type checks by type annotations
- implement new scanning probe dataclass hierarchy: `ScanConstraints` holds `ScannerAxis` and `ScannerChannel` instances, `ScanData` holds `ScanSettings` instance, `ScanData` does not need to hold complete `ScanConstraints`
- add checker and clipping methods to `ScanConstraints`
- introduce `scan_settings` attribute to `ScanningProbeInterface`
- introduce back scan data and back scan settings getter and setter methods (closes #63)
- extend constraints with respect to configurable backward scans: add new enum `BackScanCapability`

### Logic and GUI
- adapt scanning probe dummy and NI scanning probe interfuse to interface changes
- remove bool indicators (closes #64)
- refactor scanning probe logic, optimize logic and related GUI modules in preparation for configurable back scans
- improve scanning probe dummy: new image generator class, more realistic forward and backward scan simulation
- add widgets to configure back scan resolution and frequency from scanning GUI.
- add config option to save backward scan data to scanning data logic.
- improve handling of errors during start of a scan

## Motivation and Context

This refactoring was initially motivated by an extension of the scanning probe measurement toolchain to scans which are not confocal/spacial. It was decided later that this is not a good design choice. However, introducing a `ScanSettings` dataclass allowed to add backward scanning functionality (see #63) and enabled a major refactoring of the entire toolchain towards improved code readability.

## How Has This Been Tested?

This was tested with the dummy as well as nidaq hardware. There are unit tests for parts of the dataclasses.

## Types of changes

- [X] Bug fix
- [X] New feature

## Checklist:

- [X] I have documented my changes in `/docs/changelog.md`.


qku:
this addresses #90

qku:
@LukePiewalker @Neverhorst I am looking forward to your feedback!

Neverhorst (changes requested):
1. As discussed earlier, I would say decorating `ScannerChannel`, `ScannerAxis`, `ScanSettings` and `ScanConstraints` with `dataclasses.dataclass` is certainly fine.
I thought about if `ScanConstraints` is already too complex to be considered a simple data class but all the functionality implemented there is solely for read-only access of its data members and checking against them. So I guess that's fine... especially since they are all flagged as "frozen".

    However I'm not quite comfortable with `ScanData`. This object is constantly passed around different threads and even processes (i.e. via `rpyc` as remote module). In addition it serves as single source of truth for the non-volatile scan history and must be able to dump/load its data even if small changes are introduced to the class in future revisions. In my opinion there is no good way to ensure this except implementing a non-trivial dump/load functionality in the class itself (like it is done currently). All other options that come to mind would entangle qudi modules in implementation details of this data class. The decorator itself is fine, although it seems to me that there is no cleaner code obtained for this peticular case compared to a plain class definition.

2. What also needs a bit of testing is the compatibility with `rpyc` when nesting dataclasses partially as `netref`s, e.g. by the `scan_settings` property setter. This can be done via the built-in IPython console. The `numpy` array references are especially suspicious. In this case it will probably not cause problems but better safe than sorry. 

3. The type annotations are a welcome sight, but they will not work in the current form with `Python 3.8.` which we still support for qudi. For container types like `list, tuple, ...` you need to use the `typing` module instead of using the builtin types directly. I.e. replace something like:
    ```Python
    def do_stuff(arg: list[int]) -> tuple[str, ...]:
        ...
    ```
    with 
    ```Python
    from typing import List, Tuple
    
    def do_stuff(arg: List[int]) -> Tuple[str, ...]:
        ...
    ```

4. In general we should avoid using `@property` in places where non-trivial code is executed. It's ok to have a `scan_settings` read-only property but the corresponding setter will potentially execute code that leads to changes in the actual hardware. This is not the case in the current dummy and NI-card implementation but we can not assume that for other scanners. In general properties should behave like setting an instance variable and not cause a device to do something. A method/function call is much better because a programmer assumes something is happening there.

    In addition it is much more convenient (if maybe less "pythonic") when working with `Qt` because you can connect to any setter method as slot but not as easily to a property setter.

qku:
Thank you for the extensive feedback @Neverhorst.
> 1. However I'm not quite comfortable with `ScanData`. [...] In my opinion there is no good way to ensure this except implementing a non-trivial dump/load functionality in the class itself (like it is done currently). All other options that come to mind would entangle qudi modules in implementation details of this data class.

I think this is a good idea. I re-implemented `from_dict` and `to_dict` for `ScanData`.

> The decorator itself is fine, although it seems to me that there is no cleaner code obtained for this peticular case compared to a plain class definition.

I agree that the `__post__init__` method was very complicated. The reason was the following: `ScanData` needs to be constructed in two ways currently: within a hardware module using the information from a `ScanSettings` and `ScanConstraints` instance, and from a dictionary for scan history reconstruction. I now implemented a `from_constraints` class method for the former case to differentiate both cases more. Let me know if you agree that this is an improvement.

> 2. What also needs a bit of testing is the compatibility with `rpyc` when nesting dataclasses partially as `netref`s, e.g. by the `scan_settings` property setter. This can be done via the built-in IPython console. The `numpy` array references are especially suspicious. In this case it will probably not cause problems but better safe than sorry.

I successfully tested setting the scan settings of a scanner dummy from a jupyter notebook. That works fine. `numpy` arrays only occur in `ScanData` though, but at least retrieval from the dummy works fine from a notebook as well.

> 3. The type annotations are a welcome sight, but they will not work in the current form with `Python 3.8.` which we still support for qudi. For container types like `list, tuple, ...` you need to use the `typing` module instead of using the builtin types directly.

Everything should be compatible with Python 3.8 now. I tested the toolchain in a 3.8 environment.

> 4. In general we should avoid using `@property` in places where non-trivial code is executed.

I understand that as well. I left constraints as a property of the interface but changed back the settings setter to a `configure_scan` method and the `scan_data` property to a `get_scan_data`, as I already hat to implement this manually to enable connection to a Qt signal.

Neverhorst:
Thanks for the quick changes @qku ,
looks quite good. 👍

> > 2. What also needs a bit of testing is the compatibility with `rpyc` when nesting dataclasses partially as `netref`s, e.g. by the `scan_settings` property setter. This can be done via the built-in IPython console. The `numpy` array references are especially suspicious. In this case it will probably not cause problems but better safe than sorry.
> 
> I successfully tested setting the scan settings of a scanner dummy from a jupyter notebook. That works fine. `numpy` arrays only occur in `ScanData` though, but at least retrieval from the dummy works fine from a notebook as well.

What I meant by this is especially how `ScanData` fares with the GUI etc. if `ScanSettings` are created by another process via `rpyc`. The scan settings (an `rpyc` proxy object) are just added directly as reference to the `ScanData` instance.
Also the move to a shallow copy in `get_scan_data` needs to be tested because contrary to the inline comment, it contains mutable members, i.e. the `numpy` arrays. And those are heavily used by `pyqtgraph` while the hardware writes to them simultaneously. Could be working, but I'm not sure.

qku:
> What I meant by this is especially how `ScanData` fares with the GUI etc. if `ScanSettings` are created by another process via `rpyc`. The scan settings (an `rpyc` proxy object) are just added directly as reference to the `ScanData` instance.

Every time "Toggle Scan" is clicked in the GUI, the scan settings of the hardware are overwritten. The above case should not occur then if I am not mistaken?

Maybe a better test case would be having a remote scanner dummy?

> Also the move to a shallow copy in `get_scan_data` needs to be tested because contrary to the inline comment, it contains mutable members, i.e. the `numpy` arrays. And those are heavily used by `pyqtgraph` while the hardware writes to them simultaneously. Could be working, but I'm not sure.

You are right. `ScanData` needs a copy method, which copies the numpy arrays over to a new instance.

qku:
Another thing: ScanData currently has an attribute `_channel_units` which is a tuple of all channel units. This attribute is also set during instantiation. If would rather call this `channel_units` (without underscore) but that name is already taken by a property returning a dict of channel name, unit.

Is it fine to replace this property by a `get_channel_unit(channel_name) -> channel_unit`?

This is an example of a change which requires adaptation further along the toolchain (which I would take care of). Is it fine to introduce such a braking change?

Neverhorst:
That is exactly why I introduced the dict property, @LukePiewalker . It simplifies all modules that work with `ScanData`. So as long as there is no concrete reason to change that, I would like to keep it the way it is. 
You can always just iterate over `channel_units.values()` if you are only interested in the units.

For data classes I certainly have the tendency to overdo it with read-only properties because I want to make it fool-proof. I really miss private data members in Python classes. 😅

qku:
I think it is very handy to _retrieve_ units etc. via a dictionary (or a dedicated get method, but that's pretty much equal).

My idea was that _storing_ the actual units, names etc. in tuples though has the advantage that these are immutable and contain only the minimum required information. And as long as each data class contains a tuple of names to match the entries too, this should also be perfectly safe as long as you don't manually reassign the whole tuple attribute.

qku:
I was just not happy with the naming.

In `ScanConstraints`, I called the tuple containing the `ScanChannel`s `channel_objects` and the (convenience) property returning them as a dictionary with channel names as keys is still called simply `channels`.

In 'ScanData', the tuple is called `_channel_units` while the convenience dict property is called `channel_units` without underscore.

In case you agree with this way of tuple + dict retrieval: do you have an idea for a better naming convention?

qku:
> Also the move to a shallow copy in `get_scan_data` needs to be tested because contrary to the inline comment, it contains mutable members, i.e. the `numpy` arrays. And those are heavily used by `pyqtgraph` while the hardware writes to them simultaneously. Could be working, but I'm not sure.

@Neverhorst I noticed that in the way I propose to save the actual data in the data class, the numpy arrays are inside tuples. And since tuples are immutable, I believe that `ScanData` actually only contains immutable variables (plus the frozen `ScanSettings`).

Neverhorst:
> @Neverhorst I noticed that in the way I propose to save the actual data in the data class, the numpy arrays are inside tuples. And since tuples are immutable, I believe that `ScanData` actually only contains immutable variables (plus the frozen `ScanSettings`).

Unfortunately not. They are still mutable. The `tuple` itself is immutable but not the objects contained within, e.g.:
```Python
mutable_a = [1,2,3]
mutable_b = [4,5,6]
immutable = (mutable_a, mutable_b)
# Can not mutate tuple, e.g. by assigning a new item
try:
    immutable[0] = [42, 2, 3]
except TypeError:
    print('Tuple is immutable')
# However you can mutate the items themselves
immutable[0][0] = 42
assert immutable[0] == [42, 2, 3]
```

qku:
Some new thoughts from @Neverhorst : the scanning probe toolchain was designed for physically moving hardware. Adapting it for laser scans may be a bit of a stretch. Extending the functionality through inheritance may be the better approach here!

LukePiewalker:
It is one of my TODOS for a long time now but I currently barely manage to free up some time for qudi. I actually planned on testing different scenarios/ implementations to find out which might be the best and most suitable for us with regards to backwards scanning.

It kind of boils down on how one would like to configure the backwards data, where I have different scenarios in mind:

1.  Same as it has been in the "old core" by just having less points on the way back, but same clock frequency. This is kind of implemented now, yet the backwards resolution is still hard coded and not configurable unless accessing from the console.

2.  Having the same settings back and forth of scan resolution and clock frequency.

3.  Same scan resolution on the way back, but with configurable (potentially) higher clock frequency during this part of the scan.

Personally, I currently do favor option 3. Yet I still would need and like to test whether the intermediate re-configuring of the finite_samplin_io, that would be needed here, causes too much overhead.

As an additional remark: An implementation should keep the tilt correction PR #98 in mind, that this does not need too much rework.

I guess having the backwards data might be interesting for some scanning probe measurements, so I did already fantasize about whether one would like to have that in the ScanData Object as an "artificial" channel as well.
I kind of have it available with the `RawDataContainer` I once included in the interface, yet this implementation most likely will be changed in the future, such that it is directly writing to numpy arrays.

qku:
I tried out option 1 as it felt like the easiest to implement.

For this, I would suggest to add an argument to `ScanSettings`, e.g. `back_resolution`: it could default to e.g. 50 which is the case in the NI interfuse right now. Or it could default to -1, indicating that the forwards resolution should also be used for the way back (similar to option 2). In any case, I think a good default is important as many users won't care about the back scan.

Changing the frequency may be much harder if I understand the NI functionality correctly. What would be the advantage of a option 3 in your opinion @LukePiewalker ?

For getting the backwards data, I would suggest to simply duplicate the `data` property of ScanData, i.e. add a `back_data` property. Since backwards data can be acquired for every channel, this would also be a dictionary with the channels as keys.

LukePiewalker:
In favor of configurable back-scan, Yes. I think no user wants to wait on the line being scanned back as long as its regular data needs to be acquired. Using the back-scan data directly in the 2D scan is not a good idea due to "scanner drag" I've been informed once.

Retrieve back-scan data, not a must, but I thought about it. Could be a config option maybe?

Neverhorst:
In the original concept design of the new scanning probe toolchain, I certainly included fully configurable back-scans (if the hardware supports it). 
This is a feature that most commercial scanning microscopes (AFMs, confocals, etc.) provide and it has its uses. Although most of them scan the same way in both directions due to simplicity and roboustness.
Since I have just written the first part of the toolchain as a draft, this feature got lost on the way because it had no high priority.

My ideas on the topic were at the time:
1. The term back-scan refers to a full scan image. Depending on scanner hardware it could be the same scan parameters for both directions or fully configurable.
2. A back-scan (if available) should not be treated any different than a forward scan to avoid unnecessary complexity in the code. Both directions just get their own `ScanData` instance (one could be empty if no back-scan is requested). The scan settings can be set for each scan but do not differ in code structure.  
Yes, this will require changes to the interface but this is better than obscuring everything in `ScanData` or `ScanSettings`
3. Include something in the scanner constraints indicating if a backscan:
    - can be independently configured
    - is available at all
    - if it can be toggled on/off (in favor of speed)
    - has its own set of scanning constraints (might be over-the-top)
 
    The scanner constraints must basically reflect what the hardware is capable of. This should reflect the chosen option in the NI implementation @LukePiewalker.
4. It makes little sense to have a back-scan implemented and _not_ retrieve the data for it (it's free)

qku:
I like the idea of having two `ScanData` instances for the two scan directions @Neverhorst !

I guess some settings will always be the same for forward and backward scan though:
- same for forward and backward scan: axes, range, resolution of the slow axis
- (potentially) different: channels, resolution of the fast axis, frequency

To reflect this, we could include into `ScanConstraints`:
- `backscan_frequency_configurable`
- `backscan_resolution_configurable`
- `backscan_available`
- `backscan_optional`

The last two could also be merged into an enum:
```
class BackscanMode(Enum):
    UNAVAILABLE: 0
    OPTIONAL: 1
    ALWAYS: 2
```

qku:
As promised, here is my suggestion for implementing the back scan constraints:

```python
from enum import Flag, auto

class BackScan(Flag):
    """Availability and configurability of the back scan."""
    UNAVAILABLE = auto()
    SAME_AS_FORWARD = auto()
    FREQUENCY_CONFIGURABLE = auto()
    RESOLUTION_CONFIGURABLE = auto()
    FULLY_CONFIGURABLE = FREQUENCY_CONFIGURABLE & RESOLUTION_CONFIGURABLE

    CONFIGURABLE = FREQUENCY_CONFIGURABLE | RESOLUTION_CONFIGURABLE
    AVAILABLE = SAME_AS_FORWARD | CONFIGURABLE
```

One of the first five options should be selected for each hardware.

To give some examples, the following statements all evaluate to `True`:

```python
BackScan.FREQUENCY_CONFIGURABLE in BackScan.CONFIGURABLE
BackScan.FULLY_CONFIGURABLE in BackScan.CONFIGURABLE
BackScan.FULLY_CONFIGURABLE in BackScan.AVAILABLE
BackScan.FULLY_CONFIGURABLE in BackScan.FREQUENCY_CONFIGURABLE
```

I can recommend this how-to on enums: https://docs.python.org/3/howto/enum.html

Neverhorst:
Yes, that is what I meant with using `enum.Flags` @qku , thank you.

I think there are still some logical hiccups however. For example the `FULLY_CONFIGURABLE` state is ill formed. You can check this by remembering that these enums translate to integer values with certain bits set and using `bin(BackScan.FULLY_CONFIGURABLE.value)`.
It is also problematic to define the same state in opposite direction, i.e. `UNAVAILABLE` and `AVAILABLE`. What you defined with a Flag enum is basically just active bits in a bit field. You can not really define "unset bits", this needs to be checked by the user, e.g. instead of defining `UNAVAILABLE` and `AVAILABLE`, you just define one and check if it is part of the bitfield.
You should basically only define truly independent states as values in an enum and for convenience you can also define sets. 

Also the naming of the enum should be a bit more specific, i.e. instead of `BackScan` use something like `BackScanCapability`

I would propose to alter it a bit to this:

```python
class BackScanCapability(Flag):
    """Availability and configurability of the back scan."""
    AVAILABLE = auto()
    FREQUENCY_CONFIGURABLE = auto()
    RESOLUTION_CONFIGURABLE = auto()
    FULLY_CONFIGURABLE = FREQUENCY_CONFIGURABLE | RESOLUTION_CONFIGURABLE
    
# usage
capability = BackScanCapability.AVAILABLE | BackScanCapability.FREQUENCY_CONFIGURABLE

assert BackScanCapability.AVAILABLE in capability
assert BackScanCapability.FULLY_CONFIGURABLE not in capability
assert BackScanCapability.RESOLUTION_CONFIGURABLE not in capability
assert BackScanCapability.FREQUENCY_CONFIGURABLE in capability
# check for any configurability in short writing (common for bitfields)
assert BackScanCapability.FULLY_CONFIGURABLE & capability
# or a bit longer and "non-programmer-friendly"
assert (BackScanCapability.RESOLUTION_CONFIGURABLE in capability) or (BackScanCapability.FREQUENCY_CONFIGURABLE in capability)
```

In case you can not configure the back scan but it is available already implies `SAME_AS_FORWARD`. You will just get a `ScanData` object with settings you have not chosen for the back scan.

qku:
Ah! 💡I think I get it now, thanks to your explanation with the bitfields @Neverhorst . I like your suggestion and don't have any further ideas for improvement of this enum.

qku:
@timoML I am in the process of merging the tilt correction into my branch. Two thoughts:

1. I would rather not add `allow_coordinate_transform` into `ScannerConstraints`: it is an immutable now and therefore this flag cannot be overwritten by the mixin. There is an alternative though:

```python
@property
def supports_coordinate_transform(self):
    # before:
    # return self.get_constraints().allow_coordinate_transform
    # change to:
    return issubclass(self, CoordinateTransformMixin)
```

2. Make the default of `coord_transform_info` in `ScanData` an empty dict instead of `{'enabled': False}`. Whether tilt correction is enabled or not can just be found by checking whether the info dict is empty or not, right? My motivation is that setting the default value of a data class property to a non-empty dict is non-trivial.

qku:
The dummy toolchain is ready to test @LukePiewalker @timoML ! Not sure if the tilt correction works, I tried my best in resolving all merge conflicts.

The NI interfuse does not work yet.

qku:
@timoML everything should be compatible with Python 3.8 now

timoML:
From a first look with the scanning dummy, the tilt correction works as intended. Unchanged images are due to the 2D limitation of the dummy. I'll try to have a look on real hardware asap.

Concerning your two changes:
1. I find it a bit more natural to have this capability in the constraints. However, your simplification is probably worth dropping it from the constraints given the lower code complexity.
2. I would rather not derive the "enabled" flag implicitly from an empty dict. Could we maybe?

  ```
  def _get_coord_transform_info(self) -> Dict[str, Any]:
        rad_2_deg = lambda phi: phi / np.pi * 180
        info = {'enabled': False}
        if self.coordinate_transform_enabled:
            info.update({
                'enabled': True,
                'transform_matrix': self._coordinate_transform_matrix.matrix,
                'tilt_angle (deg)': rad_2_deg(self._calc_matr_2_tiltangle()),
                'translation': self._coordinate_transform_matrix.matrix[0:3, -1]
             })
```
       
This way: 
- hw with supports_coordinate_transform==False: empty dict
- hw supports_coordinate_transform==False but turned off: {'enabled':False}

**Comment by timoML** [comment]
@qku 

It looks like the backscan is not implemented for the `ni_scanning_probe_interfuse` yet.
Is this still on your todo list?

 `Can't instantiate abstract class "NiScanningProbeInterfuse" with abstract methods {'configure_back_scan', 'back_scan_settings'}`

**Comment by qku** [comment]
Yes, that's correct! It's the last bigger thing that needs to be done. I think I can do it this week.

File: src/qudi/hardware/interfuse/ni_scanning_probe_interfuse.py [hunk 1]
```diff
@@ -429,8 +430,10 @@ def stop_scan(self):
         """
         #self.log.debug("Stopping scan")
         if self.thread() is not QtCore.QThread.currentThread():
-            QtCore.QMetaObject.invokeMethod(self, '_stop_scan',
-                                            QtCore.Qt.BlockingQueuedConnection)
+            #QtCore.QMetaObject.invokeMethod(self, '_stop_scan',
+            #                                QtCore.Qt.BlockingQueuedConnection)
+            # FIXME: why does the above not work anymore?
+            self._stop_scan()
```
qku, line 436: Not sure why, but somehow `invokeMethod` does not work anymore if the tilt correction mixin is used. It complains that there is no such method on `NIScanningProbeInterfuse`. It works if only the bare interfuse is loaded.

qku:
Other than the issue with `invokeMethod`, configurable back scans work with the NI interfuse now. I also tested it with real hardware.

src/qudi/hardware/interfuse/ni_scanning_probe_interfuse.py
Neverhorst: `invokeMethod` works in mysterious ways in `PySide2`. One of those mysteries is, that it only works if the slot in question has been decorated with the `PySide2.QtCore.Slot` decorator.

It's one of many reasons why using signals is preferred over `invokeMethod`.

qku: I was able to resolve this issue by re-implementing `_start_scan` and `_stop_scan` in the class which inherits the bare scanner and the tilt correction mixin.

qku:
Did anybody have a chance to test this?

timoML:
Thanks for your (huge) effort. This looks gread overall!
I tested it on a nicard + PI xyz scanner setup:
- saving back scan data worked flawlessly
- tilt correction worked as expected

There are some smaller points from the user perspective I noticed:

**StatuVars**
When checking out from current main, the loading of the old statusVars breaks at scanning_data_logic::__scan_history_from_dicts().
The incompatible StatusVar should be handled, most likely by cleaning all scan history entries with the old format.
Otherwise users will have a hard time updating.

```
Error while settings status variables:
  File "C:\Users\Setup3-PC\Desktop\qudi_newrepo\qudi-core\src\qudi\core\module.py", line 338, in _load_status_variables
    value = var.constructor_function(self, value)
  File "C:\Users\Setup3-PC\Desktop\qudi_newrepo\qudi-iqo-modules\src\qudi\logic\scanning_data_logic.py", line 116, in __scan_history_from_dicts
    for data_dict, back_data_dict in history_dicts:
ValueError: too many values to unpack (expected 2)
```

**The button** to configure back scans in the GUI/Axes control seems un-intuitive or buggy:
- default behavior, when coming with no StatusVars for the "=" checkbox should be to be enabled. Users without experience in backscan will be overwhelmed otherwise.
- Not sure how it was before, but default scan resolutions with no StatusVars is 1 px for me. Should better be something like 100 px. 
- The same for scan frequencies. For me, they default to 1 Hz. (Better: 500 Hz).
- The following yields an error for me:
1. Uncheck "=" buttons for X/Y
2. Set new values for resolution X: 70 / 70, Y: 70 / 70
3. Change Y to Y: 70 / 100
4. Check "=" button for X
5. Start scan

```
 File "C:\Users\Setup3-PC\Desktop\qudi_newrepo\qudi-iqo-modules\src\qudi\logic\scanning_probe_logic.py", line 512, in start_scan
    self._scanner().configure_back_scan(back_settings)
  File "C:\Users\Setup3-PC\Desktop\qudi_newrepo\qudi-iqo-modules\src\qudi\hardware\interfuse\ni_scanning_probe_interfuse.py", line 281, in configure_back_scan
    self.constraints.check_back_scan_settings(settings, forward_settings)
  File "C:\Users\Setup3-PC\Desktop\qudi_newrepo\qudi-iqo-modules\src\qudi\interface\scanning_probe_interface.py", line 181, in check_back_scan_settings
    raise ValueError('The back scan must use the same slow axis resolution as the forward scan.')
ValueError: The back scan must use the same slow axis resolution as the forward scan.

```

qku:
Thank you very much for the detailed feedback @timoML ! I am glad that overall everything works for you. I addressed your feedback in the last four commits. I hope the commit messages speak for themselves. Let me know if you find anything else!

qku:
@timoML i tested the current status on a confocal setup with NI card. Everything is working as expected.

timoML (dismissed):
I tested it on my setup with a nicard; works fine. Thank you for this big contribution.

Neverhorst (approved):

qku:
@Neverhorst @timoML I merged main again into this branch, resolving the conflicts with the new custom saving nametag. It is ready to merge from my side.

timoML (approved):
Thanks again, looks good.

PR #98, opened 2023-09-12
Tilt correction 

timoML:
This PR re-introduces the tilt correction known from the old core. 

## Description
The tilt correction now works as mixin class that can be added to any hardware implementations of the scanning_probe_interface. The mixin introduces a virtual coordinate system that can be transformed by an arbitrary transformation matrix. For realizing a tilt correction in the scanning gui, support vectors can be selected from which a transformation matrix onto the e_z plane is calculated and fed to the mixin.

Note that for this PR to work, the core branch tilt_correction_main has to be checked out as well.

Thanks at the project team @Geegee, @anjusha-vs, @rbs220896 and eveyone who tested (in particular @NicolasStaudenmaier). 

## How Has This Been Tested?
Setup3, dummy, setup of @NicolasStaudenmaier.

## Screenshots:
grafik

## Types of changes
- [ ] Bug fix
- [x] New feature

## Checklist:


timoML:
Up for discussion:
- Right now, users have to change their config to enable tilt correction for the NiScanningProbeInterfuse and the scanner_dummy and set their hw to the new mixed-in class (eg. [NiScanningProbeInterfuseCorrected](https://github.com/Ulm-IQO/qudi-iqo-modules/blob/df7882d754f7fdec1c2fdef908d5e681c8bf52c7/src/qudi/hardware/interfuse/ni_scanning_probe_interfuse.py#L932)). I think one could simplify this by rename the mixed-in classes to the current bare classes. As long as no correction matrix is actively configured by the user, the behavior won't change.
- I added the helper method [_expand_coordinate()](https://github.com/Ulm-IQO/qudi-iqo-modules/blob/df7882d754f7fdec1c2fdef908d5e681c8bf52c7/src/qudi/interface/scanning_probe_interface.py#L158) to the ScanningProbeInterface. Strictly speaking, this function shouldn't live in an interface class and one would create a util class that all implementations of the interface can access. For a single function, this seems a too big code overhead, though.

prithviulm:
Some bug with status variables that can be fixed with changing one line in the `scanning_probe_interface.py` at line 408 from
`if dict_repr['coord_transform_info'] is not None:` to `if dict_repr.get('coord_transform_info') is not None:`.

LukePiewalker:
As discussed in today's meeting, we should check what happens, if the the tilt correction produces value out of the scan range. We should avoid a "kink" in the real scan data or a silent clipping.

timoML:
As discussed I resolved the issue. The ni_scanner_interfuse will now issue
 
     ValueError: Scan axis z out of range [-0.0001, 0.0001]

on configuration of an out-of-range scan.

@Neverhorst, @LukePiewalker, @qku 
Because the ScanningProbeInterface is not aware of the scan grid, I don't see how to handle this in the CoordinateTransformMixin. For now, handling in on the hw level should be ok.

timoML:
I think we're good to merge. 
@Neverhorst @LukePiewalker 
Can you approve?

Issue #88, opened 2023-08-28
[Bug] saving ORMD fitting data


Maabur:
### Version

current main

### What is affected by the bug?

Saving CWODMR fitting data spectra 

### When does the bug occur?

After acquiring a CW spectra and fitting wit a gaussian dip. The data saved has 3 png files two of them contain some mV plots versus the microwave frequency (see enclosed). But I cannot find the fit result data such as the center frequency and the fwhm

### How do we replicate the issue?

Acquire a CW odmr spectra, perform a fit and save.

### Expected behavior

I expect to find somewhere the fitting results data at least on the png file or on the .dat file

### Relevant log output

```shell
none
```

### Additional Comments

![20230828-1719-06_test_ODMR_ai0_range0](https://github.com/Ulm-IQO/qudi-iqo-modules/assets/75795550/06d83f1d-b2fb-4c10-90bf-27859afd99f8)
![20230828-1719-06_test_ODMR_ai1_range0](https://github.com/Ulm-IQO/qudi-iqo-modules/assets/75795550/1ef1611e-460a-4f82-a65c-45d37593b764)
![20230828-1719-06_test_ODMR_pfi3_range0](https://github.com/Ulm-IQO/qudi-iqo-modules/assets/75795550/812cc002-0890-410f-9926-3c501dc491b5)

timoML:
Merged with #99.

PR #99, opened 2023-09-13
Add fit parameters to metadata for saved files (pulsed, ODMR)

NicolasStaudenmaier:
Fit parameters in `pulsed` and `ODMR` are now saved in the metadata of data files when saving a measurement. 

## Description

Adding the fit model name and the fit parameters according to the formatted result output. Depending on the availability of a fit for the (alternative) data in pulsed or on the channel and range in ODMR the data is added. If no fit is available, no fit data is added to the metadata. 
In pulsed the fit is saved in the pulsed_measurement data file, and in ODMR in all text files. 

## Motivation and Context

In pulsed the fit parameters are only printed next to the figure in the png file. For ODMR no fit parameters are saved at all. It might be useful to access the fit parameters from the text file. 

Solves issue #88. 

## How Has This Been Tested?

Tested with dummy configuration of the current version. 

## Types of changes

- [x] Bug fix
- [ ] New feature

## Checklist:


takuya-ulm (commented):
I got \n in the fit result as follows:
# fit result (channel "apd counts" range 0)='Lorentzian Dip: offset: 200620.0 ± 60.0  \ncenter: 2865.1 ± 0.4 M \nsigma: 10.0 ± 0.6 M \namplitude: -6.8 ± 0.28 k \n'
 This was originated from create_formatted_output in qudi.util.units.
I want to suggest removing them. For example, like shown in the code.

timoML:
I looked into .summary(). Looks like this dict would contain what we need in ['params']. However, it was only introduced in lmfit 1.2.0 (5/2023) and would require a change in our setup requirements. For now, I introduced a qudi-core helper [here](https://github.com/Ulm-IQO/qudi-core/tree/save_fit_in_metadata).

Neverhorst:
> I looked into .summary(). Looks like this dict would contain what we need in ['params']. However, it was only introduced in lmfit 1.2.0 (5/2023) and would require a change in our setup requirements. For now, I introduced a qudi-core helper [here](https://github.com/Ulm-IQO/qudi-core/tree/save_fit_in_metadata).

If the modules in the `qudi-iqo-modules` addon package are using this feature, it would make sense to just include this package version requirement in its `setup.py`, e.g. `lmfit>=1.2.0`. Problem solved and no need to clutter `qudi-core` with something that is outdated anyway. 👍

timoML:
Unfortunately, it is more complicated than I suspected: lmfit's .summary() returns the parameters as a tuple (not a dict) with hard coded order. When dumping, this yields human unreadable results like:

`'params': [('offset', 200371.95891243816, True, None, 186042.13771090773, 208322.72278583102, None, 123.92946774026674, {'center': 2.0450782521128066e-06, 'sigma': 0.2311852154836438, 'amplitude': -0.1359148619646129}, 200074.4790258307, None), ('center', 2867441428.334397, True, None, 2770000000.0, 2970000000.0, None, 308997.90197928884, {'offset': 2.0450782521128087e-06, 'sigma': -1.2424267099560231e-05, 'amplitude': -2.0366472860784767e-05}, 2868000000.0, None), ('sigma', 2751341.5731732002, True, None, 1000000.0, 100000000.0, None, 317605.2799702476, {'offset': 0.23118521548364368, 'center': -1.2424267099560231e-05, 'amplitude': 0.5250577474849855}, 1944560.4652284337, None), ('amplitude', -6567.367922722109, True, None, -6567.367922731617, 0, None, 64.77947751158845, {'offset': -0.13591486196461305, 'center': -2.0366472860784767e-05, 'sigma': 0.5250577474849853}, -3283.6839613658085, None)]}`

while we would prefer a human readable, dict-like style (as with the added core .dict_result() function):
`'offset': {'value': 200917.5114554685, 'min': 188589.88295235785, 'max': 207777.30558512648, 'stderr': 115.82321083006927}, 'center': {'value': 2869935442.325448, 'min': 2770000000.0, 'max': 2970000000.0, 'stderr': 521566.4696499122}, 'sigma': {'value': 4719809.358276054, 'min': 1000000.0, 'max': 100000000.0, 'stderr': 549534.733591269}, 'amplitude': {'value': -4465.000169262333, 'min': -6501.286534228013, 'max': 0, 'stderr': 435.03743948837626}}`

This means that also with .summary(), we need a helper that makes it human readable.

@Neverhorst 
I implemented both ways [here](https://github.com/Ulm-IQO/qudi-core/compare/main...save_fit_in_metadata#diff-4d64131d4d2605ef62da7af13b8f58b34a9694b6a0e98ac19380c778cb4385a5). What do you prefer:
1. New style with .dict_result() helper in qudi-core. Will require bumping lmfit to 1.2.0 in qudi-core.
2. As 1, but we move the helper (a bit artifically) to qudi-iqo-modules
3. Old style

Neverhorst:
@timoML , ok I see now, it is not so simple.

While looking into the matter I realised however that we should probably agree on what information we want to be included in the file header of each data file.
In your example you also save the boundary conditions (`min` and `max`) together with the fit result (`value` and `stderr`). This seems a bit arbitrary to me because you combine fit result with fit model information but only part of it.
For the sake of consistency I would propose _either_ saving the entire information contained in `lmfit.ModelResult` _or_ just the parameter results, i.e. `value +- stderr`. The first will not be human readable and needs tooling to extract the full `ModelResult` object from a data file while the latter can be quite easily implemented and is human readable (my favourite).

This question again reiterates the old conflict of scientists wanting to hoard every shred of information in this single file (even though it is never accessed in 99.9% of the time) vs. easy use and readability. From experience I would strongly advocate the path of readability. The new way of fitting in `qudi` allows to perform the fit again any time without starting `qudi` if the need arises (e.g. you need the fit statistics/metrics for a paper). So as long as the file header contains enough information to reproduce the fit, I see no reason to clutter it with meaningless information that defeats the benefits of a plain text, human-readable file in the first place. If you want maximum information, you create proper databases and work with queries/tooling to retrieve pieces of information and do not use text files.

TL;DR:
I would propose the fit information dict should look like:
```python
{'offset' : '(200918 +- 116)c/s', 
 'center': '(2.8699 +- 0.0006)GHz',
 ...}
```
And of course the metadata should also contain the fit type.

timoML:
@Neverhorst 
I think there is currently no general way of saving the units with arbitrary fit parameters. Eg. take the dimension-less parameter beta in streched decays. Of course the x/y axis units are stored in the meta data property 'column_headers'.

Neverhorst:
> @Neverhorst I think there is currently no general way of saving the units with arbitrary fit parameters. Eg. take the dimension-less parameter beta in streched decays. Of course the x/y axis units are stored in the meta data property 'column_headers'.

Have a look at how I implemented that [here](https://github.com/Ulm-IQO/qudi-core/compare/formatted-fit-result-dict). Pretty much in the same way as `FitContainer.formatted_result`. The logic can pass units to the helper function. If none are given, none are formatted into the string.
Independent on how the result is formatted in the end (string vs. dict) I would prefer this structure because it is in line with the current way of handling results (all in the same place and works the same way).

timoML:
I see, I like the mechanism. (Although I think our current logics would hardly know the units for arbitrary fit functions.)

Btw: What's the reasoning in dropping fixed parameters in [formatted_result](https://github.com/Ulm-IQO/qudi-core/blob/7e6c5db57d67089b96ad6f81229642900c7a763d/src/qudi/util/datafitting.py#L345C12-L345C31)? For me it's a bit unexpected that a fixed parameters won't show eg. in the pulsed gui fit result.

Neverhorst:
> I see, I like the mechanism. (Although I think our current logics would hardly know the units for arbitrary fit functions.)

True, but at least you have a mechanism where this _could_ be customized in the future.

> Btw: What's the reasoning in dropping fixed parameters in [formatted_result](https://github.com/Ulm-IQO/qudi-core/blob/7e6c5db57d67089b96ad6f81229642900c7a763d/src/qudi/util/datafitting.py#L345C12-L345C31)? For me it's a bit unexpected that a fixed parameters won't show eg. in the pulsed gui fit result.

I think this was a quick fix because non-varying parameters have no `stderr` set. Also it's a matter of taste because some simplified fit models are derived from more complex models with some parameters being fixed (e.g. offset) and having them around would give a false impression on model complexity.
But yeah, I also understand the desire to have all shown. We could include them if you like but we should then make it a bit more complex by flagging the fixed parameters as such in the output string.

timoML:
After our internal coordination, I implemented the dict solution [here](https://github.com/Ulm-IQO/qudi-core/compare/main...save_fit_in_metadata).
I also intruduced a distinction mechnasim between fixed parameters 
image
and unsucessfull fits:
image

@Neverhorst 
If you approved, I can continue with the PR in core before merging this PR here.

timoML (approved):
All questions resolved, thank you.

PR #109, opened 2023-11-08
wavemeter: don't stop measurement

qku:
## Description

The HighFinesse wavemeter has exposed start/stop measurement functions. They start/stop the hardware-internal measurement loop. If this is not running, no new values are acquired.

This PR changes the qudi behavior to make sure to start the measurement but not to bother about stopping it once no qudi module is active anymore.

It also adds a check to properly catch an error code which indicates that the wavemeter app is not running and passes it to the user as an exception.

## Motivation and Context

In this way, other processes that need wavemeter data are not affected. The most prominent example is the built-in laser stabilization of the wavemeter application as pointed out by @Nick-Grimm .

## How Has This Been Tested?

This was tested with a HighFinesse WS 7. Special testing with remote modules should not be necessary.

## Types of changes

- [X] Bug fix
- [ ] New feature

## Checklist:

### DIFF
--- a/src/qudi/hardware/wavemeter/high_finesse_proxy.py
+++ b/src/qudi/hardware/wavemeter/high_finesse_proxy.py
@@ -133,8 +133,6 @@ class HighFinesseProxy(Base):
         else:
             self._wm_has_switch = False
 
-        self._stop_measurement()
-
         self._set_up_watchdog()
 
     def on_deactivate(self) -> None:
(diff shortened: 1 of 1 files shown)


qku:
@Nick-Grimm does this solve your issue? Can you give the changes a test?

Nick-Grimm:
Hi @qku, 
I tested the changes and they work perfectly fine. Therefore, I would consider these small changes ready to merge. 
Thank you!

Nick-Grimm (approved):
Approved 👍

PR #118, opened 2023-12-11
improved support for SRS signal generators

qku:
## Description

This PR introduces improvements to the microwave hardware file for Stanford Research Systems (SRS) signal generators.

- support for all current SRS signal generators (see [manufacturer website](https://www.thinksrs.com/products/test.html))
- proper matching of the model identifier to supported functions and hardware constraints
- check for error after every write command
- display current frequency during scan
- type annotations and documentation

## Motivation and Context

We are using a non-vector SRS signal generator in our lab. It was not compatible with the current hardware file due to the usage of commands which are defined exclusively for vector models.

## How Has This Been Tested?

Tested on real hardware (SG384) with the ODMR toolchain.

## Types of changes

- [ ] Bug fix
- [X] New feature

## Checklist:

- [X] I have documented my changes in `/docs/changelog.md`.

### DIFF
--- a/src/qudi/hardware/microwave/mw_source_srssg.py
+++ b/src/qudi/hardware/microwave/mw_source_srssg.py
@@ -69,13 +69,14 @@ class MicrowaveSRSSG(MicrowaveInterface):
         self._rm = visa.ResourceManager()
         self._device: MessageBasedResource = self._rm.open_resource(  # type: ignore
             self._visa_address,
-            timeout=int(self._comm_timeout * 1000)
+            timeout=int(self._comm_timeout * 1000),
+            read_termination='\r\n'
         )
 
         # Reset device
-        self._device.write('*RST')
-        self._device.write('ENBR 0')  # turn off Type N output
-        self._device.write('ENBL 0')  # turn off BNC output
+        self._write('*RST')
+        self._write('ENBR 0')  # turn off Type N output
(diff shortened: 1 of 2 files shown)


src/qudi/hardware/microwave/mw_source_srssg.py
```diff
@@ -21,10 +25,8 @@
 """
 
 
-try:
-    import pyvisa as visa
-except ImportError:
-    import visa
+import pyvisa as visa
+from pyvisa.resources import MessageBasedResource
```
NicolasStaudenmaier, line 29: Is there any particular reason why you changed the imports? I would suggest
```
try:
    import pyvisa as visa
    from pyvisa.resources import MessageBasedResource
except ImportError:
    import visa
    from visa.resources import MessageBasedResource
```

NicolasStaudenmaier (commented):
I quickly tested with a SRS384 and it seems to work. I have only tested for possible errors, not for correct functionality. 
Code review looks good to me, except for one comment at the imports. If we make the visa import consistent I approve the PR.

qku:
Thank you for reviewing @NicolasStaudenmaier ! I changed the visa imports as suggested.

NicolasStaudenmaier (approved):
Looks good. Thanks Kilian for the work on it.

Issue #117, opened 2023-11-28
[Bug] Missing/wrong generation_method_parameters for PulseSequences


Nick-Grimm:
### Version

current main

### What is affected by the bug?

In the pulsed toolchain any uploaded `PulseSequence` from predifined methods does not show its own `generation_method_parameters` but rather the parameters of a former uploaded 'PulseBlockEnsemble'. This leads to false saving of the generation method paramters in the Metadata.

### When does the bug occur?

The bug occurs when uploading predefined methods written as PulseSequences, e.g. `t1_seq`.

### How do we replicate the issue?

-Upload any predefined method written as PulseSequence, e.g. `t1_seq`
-Check in the console the `pulsed_measurement_logic.generation_method_parameters`

### Expected behavior

The generation method parameters should match. 

### Additional Comments

I think the issue is resolved when adding for sequences in the `sequence_generator_logic` in the method `generate_predefined_sequence`: `sequence.generation_method_parameters = kwargs_dict`  just as for ensembles.

PR #120, opened 2024-01-23
Laser scanning toolchain used for e.g. PLE 

Nick-Grimm:
## Description

This PR implements the functionality of the laser scanning toolchain `(wavemeterlogger)` used for PLE scanning of the old core in the new core. It is based on the new wavemeter data instream hardware file developed by @qku that was merged with PR #79 and started with discussion in #59. 

It also contains the solution for bug #117.

## Motivation and Context

The heart of this toolchain is to synchronize counts received through the independent `time_series_reader_logic` with the queried wavelength values received from the wavemeter instream hardware device and displaying this with a histogram. For this, it uses software based timing such that the `wavemeter_scanning_logic_3` is connected to a QT signal from the `time_series_reader_logic` delivering the counts. Upon this rate the wavemeter values are queried and interpolated onto the values from the `time_series_reader_logic`. 
Further it provides the ability to additionally display the wavemeter values over time or solely to display the wavemeter values. 

## How Has This Been Tested?

It is tested and used extensively in laboratory with a remote HighFinesse WS7 wavemeter and a NI X instreamer. Further it has been tested with two `data_instream_dummy` instances (for one sample generation is matched for wavemeter values).

## Types of changes

- [ ] Bug fix
- [x] New feature

## Checklist:


src/qudi/hardware/wavemeter/high_finesse_wavemeter.py
qku, line 1: I believe `read_data_into_buffer` should be blocking until the all requested samples are available (see the docstring in the interface). Why did you add this timeout? Can you have this timeout in your logic instead?

src/qudi/hardware/dummy/data_instream_dummy_wavemeter.py
qku, line 1: This dummy is different from the `data_instream_dummy` only in minor details:
- it uses a different default sample rate
- `_make_counts_func` uses different scaling
I suggest to add one or more config options to `data_instream_dummy` to make it more flexible instead of adding a new dummy.

src/qudi/wavemter_rework.cfg
qku, line 1: This could be embedded into a documentation page on setting up this toolchain, what do you think?

qku (changes requested):

src/qudi/hardware/wavemeter/high_finesse_wavemeter.py
Nick-Grimm: I have added this timeout because if the wavemeter is either under- or overexposed the function is blocking for long time until the requested samples are available. Might make more sense to handle this in the logic but I was thinking this would concern other logic modules as well. What do you think?

src/qudi/wavemter_rework.cfg
Nick-Grimm: Yes definitely, at latest when the PR is ready for merge.

src/qudi/hardware/wavemeter/high_finesse_wavemeter.py
qku: Interesting. The hardware is supposed to give NaN values during under- or overexposure. But that should not delay anything.

Could it be that you are working with automatic exposure and that the wavemeter is going to really high exposure time values upon loss of signal?

Neverhorst: Well, this is exactly why you also have `DataInStreamInterface.read_available_data_into_buffer` which only reads data that is available without waiting. It does not make much sense to me to add a timeout to the very function that is supposed to wait for all samples. This smells more like a design flaw in the logic (do not request more samples than you want to wait for, or handle it dynamically with the afore mentioned method) or the hardware module (better handling of under-/over-exposure).

Nick-Grimm: > Could it be that you are working with automatic exposure and that the wavemeter is going to really high exposure time values upon loss of signal?

Yes, we are working in auto exposure mode and that is exactly the issue.

PR #121, opened 2024-02-02
Allow for inverted trigger polarity of the NI card

JohannesWes:
feat: enable changing trigger polarity

add config flag invert_trigger_polarity. If invert_trigger_polarity, set the idle_state of the NI card to HIGH (instead of LOW) and use FALLING edges instead of RISING edges.

If both HIGH and LOW active devices are used, the change could lead to problems.


NicolasStaudenmaier:
Hi @JohannesWes, thank you for opening this Pull Request. The forked branch that you would like to merge here contains a lot of different and custom changes. I would like to ask you to remove them. The easiest would be to create a new branch from the current version of the main branch, commit your changes there and make a new pull request with that. Then we can review the changes properly. It's also be best to fill in the predefined suggestions for the description of the PR.

PR #89, opened 2023-08-29
Hotfix poimanagerlogic crash without image loaded 

timoML:
One liner that fixes #87. This would crash with 

```
TypeError: cannot unpack non-iterable NoneType object
 in save_roi()
    x_extent, y_extent = self.roi_scan_image_extent
```

## How Has This Been Tested?
Tested on dummy.

## Types of changes
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change (Causes existing functionality to not work as expected)

## Checklist:
- [x] My code follows the code style of this project.
- [ ] I have documented my changes in `/docs/changelog.md`.
- [ ] My change requires additional/updated documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added/updated the config example for any module docstrings as necessary.
- [x] I have checked that the change does not contain obvious errors
(syntax, indentation, mutable default values, etc.).
- [ ] I have tested my changes using 'Load all modules' on the default dummy configuration.
- [x] All changed Jupyter notebooks have been stripped of their output cells.

PR #125, opened 2024-04-08
Updating gitignore

prithviulm:
## Description
Updating the `.gitignore` file to be more relevant.

## Motivation and Context
The `.gitignore` file misses several files that new users may accidentally commit to the repo. Also various directories and cache files generated by testing suites, documentation generation, other IDEs (e.g. VS Code) are not currently excluded by the `.gitignore` file.

## How Has This Been Tested?
I have only tested this locally with the latest commit 212103937862ffe0744661045004858d728ad173 from the main branch and checked that no important core files or others have been accidentally excluded.

## Types of changes
- [x] Bug fix
- [ ] New feature

## Checklist:


timoML:
We should definitely keep `.idea` in as it's the Pycharm config folder.

I didn't go through all the things you dropped, but I'm not sure on:
-  '.DS_store' as it is Mac specific.
- .swp seems to be vim related
I think we should only drop things that we are sure won't break for other users.
(Eg. the things related to c compiler toolchains should be safe)

Can you double check that we can re-distrubute code licensed under CC0 with a new license (LGPL)?

prithviulm:
@timoML I've made the suggested changes.

timoML (approved):
Thanks (see also qudi-core PR #89).

PR #129, opened 2024-05-08
Introducing cursor position in Scan2DWidget

Geegee:
There is now another label in the Scan2DWidget showing you 
the position of the mouse cursor.

## Motivation and Context

It is additional information allowing one to get the exact position of e.g. a colorcenter
without actually moving the scanner.

## How Has This Been Tested?

Only on dummy for now

image

## Types of changes

- [ ] Bug fix
- [x] New feature

We should probably check performance across different setups
and find a sensible default for the polling rate of the cursor position.


Neverhorst (approved):
Just fixed some minor thing and added axis units to the position display. Also made the max update rate an optional `__init__` parameter to be customized in some other scenario.
Looks good to me now.

PR #131, opened 2024-05-13
Spectrum instrumentation ADC as fast_counter hardware

takuya-ulm:
## Description

AD converter from spectrum instrumentation is implemented as fast_counter hardware.

## Motivation and Context

AD card from spectrum instrumentation is newly implemented.
The class structure is organized for robustness.

## How Has This Been Tested?

This hardware is tested in the pulsed toolchain.
This change does not affect the rest of the code.

## Types of changes

- [ ] Bug fix
- [x] New feature

## Checklist:

### DIFF
--- a/src/qudi/gui/time_series/time_series_gui.py
+++ b/src/qudi/gui/time_series/time_series_gui.py
@@ -82,6 +82,8 @@ class TimeSeriesGui(GuiBase):
 
         self._channels_per_axis = [set(), set()]
 
+        self._single_axis = True
+
     def on_activate(self):
         """ Initialisation of the GUI """
         self._mw = TimeSeriesGuiMainWindow()
@@ -334,15 +336,15 @@ class TimeSeriesGui(GuiBase):
             self._mw.trace_plot_widget.hideAxis('right')
             self._channels_per_axis = [tuple(enabled), tuple()]
             self._mw.trace_plot_widget.setLabel('left', 'Signal', units='')
-        elif len(enabled) == 2:
-            self._mw.trace_plot_widget.hideAxis('right')
-            self._channels_per_axis = [(enabled[0],), (enabled[1],)]
-            self._mw.trace_plot_widget.setLabel('left',
-                                                enabled[0],
-                                                units=channel_units[enabled[0]])
-            self._mw.trace_plot_widget.setLabel('right',
-                                                enabled[1],
-                                                units=channel_units[enabled[1]])
+        # elif len(enabled) == 2:
+        #     self._mw.trace_plot_widget.hideAxis('right')
+        #     self._channels_per_axis = [(enabled[0],), (enabled[1],)]
+        #     self._mw.trace_plot_widget.setLabel('left',
+        #                                         enabled[0],
+        #                                         units=channel_units[enabled[0]])
+        #     self._mw.trace_plot_widget.setLabel('right',
+        #                                         enabled[1],
+        #                                         units=channel_units[enabled[1]])
         elif len(enabled) > 2:
             self._mw.trace_plot_widget.hideAxis('right')
             self._channels_per_axis = [tuple(enabled), tuple()]
--- a/src/qudi/hardware/fast_adc/spectrum_instrumentation/si_commander.py
+++ b/src/qudi/hardware/fast_adc/spectrum_instrumentation/si_commander.py
@@ -100,7 +100,7 @@ class Commander:
         """
 
         trig_reps = self.cmd.process.get_trig_reps()
-        processed_reps = self.processor.avg.num
+        processed_reps = self.processor.data.avg.num_rep
         unprocessed_reps = trig_reps - processed_reps
         if self._reps != 0 and processed_reps >= self._reps:
             return
--- a/src/qudi/hardware/fast_adc/spectrum_instrumentation/si_commander.py
+++ b/src/qudi/hardware/fast_adc/spectrum_instrumentation/si_commander.py
@@ -1,0 +1,108 @@
+# -*- coding: utf-8 -*-
+
+"""
+This file contains the data classes for spectrum instrumentation fast counting devices.
+
+Qudi is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+Qudi is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with Qudi. If not, see <http://www.gnu.org/licenses/>.
+
+Copyright (c) the Qudi Developers. See the COPYRIGHT.txt file at the
+top-level directory of this distribution and at <https://github.com/Ulm-IQO/qudi/>
+"""
+from qudi.hardware.fast_adc.spectrum_instrumentation.si_settings import MeasurementSettings
+
+class Commander:
+    def __init__(self, cmd):
+        self.cmd = cmd
+        self.processor = None
+
+        self._seq_size_B = 0
+        self._reps = 0
+        self.unprocessed_reps_limit = 0
+
+    def init(self, ms: MeasurementSettings, data_processor):
+        self._seq_size_B = ms.seq_size_B
+        self._reps = ms.reps
+        self.unprocessed_reps_limit = ms.reps_per_buf
+
+        self._assign_process(ms.gated, ms.data_stack_on)
+
+        self.processor = data_processor
+
+    def _assign_process(self, gated, stack_on):
+
+        def process_data_ungated_stackoff():
+            curr_avail_reps = self.cmd.process.get_curr_avail_reps()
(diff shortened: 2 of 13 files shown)


src/qudi/gui/time_series/time_series_gui.py
TobiasSpohn, line 1: For a new hardware implementation this file should not be changed.

src/qudi/hardware/fast_adc/spectrum/si_commands/buffer_commands.py
```diff
@@ -0,0 +1,127 @@
...
+either version 3 of the License, or (at your option) any later version.
+
+Qudi is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+See the GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License along with qudi.
+If not, see <https://www.gnu.org/licenses/>.
+"""
+from pyspcm import *
```
TobiasSpohn, line 22: Import only the function and classes that are needed in the file, not the whole package.

src/qudi/hardware/fast_adc/spectrum/si_commands/card_commands.py
```diff
@@ -0,0 +1,75 @@
...
+either version 3 of the License, or (at your option) any later version.
+
+Qudi is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+See the GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License along with qudi.
+If not, see <https://www.gnu.org/licenses/>.
+"""
+from pyspcm import *
```
TobiasSpohn, line 22: Import only the function and classes that are needed in the file, not the whole package.

src/qudi/hardware/fast_adc/spectrum/si_commands/configure_commands.py
```diff
@@ -0,0 +1,549 @@
...
+Qudi is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+See the GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License along with qudi.
+If not, see <https://www.gnu.org/licenses/>.
+"""
+import ctypes
+from functools import reduce
+from pyspcm import *
```
TobiasSpohn, line 24: Import only the function and classes that are needed in the file, not the whole package.

src/qudi/hardware/fast_adc/spectrum/si_commands/si_commands.py
```diff
@@ -0,0 +1,128 @@
...
+        curr_avail_reps = self.get_curr_avail_reps()
+        while curr_avail_reps == 0:
+            curr_avail_reps = self.get_curr_avail_reps()
+            current_time = time.time() - initial_time
+            if current_time > 10:
+                time_out = True
+                return time_out
+            time.sleep(self.wait_time_interval)
+        time_out = False
+        return time_out
```
TobiasSpohn, line 128: This function is never used

src/qudi/hardware/fast_adc/spectrum/si_main.py
```diff
@@ -0,0 +1,411 @@
...
+
+You should have received a copy of the GNU Lesser General Public License along with qudi.
+If not, see <https://www.gnu.org/licenses/>.
+"""
+import time
+import numpy as np
+
+from enum import IntEnum
+from pyspcm import *
+from spcm_tools import *
```
TobiasSpohn, line 26: excplicitly import only the necessary classes and functions, not the whole module.

src/qudi/hardware/fast_adc/spectrum/si_main.py
```diff
@@ -0,0 +1,411 @@
...
+        Receives the current status of the Fast Counter and outputs it as
+                    return value.
+
+                0 = unconfigured
+                1 = idle
+                2 = running
+                3 = paused
+                -1 = error state
+        """
+        return self._internal_status
```
TobiasSpohn, line 293: As far as I understood, the _internal_status is just a python variable, not associated with any response from the hardware itself. Thus during a measurement an error could occur with the card but when checking with get_status() one would still get the result that the card is running as _internal_status was set to this during the start of the measurement.
Maybe the interface allows for a status query of the card?

src/qudi/hardware/fast_adc/spectrum/si_commands/si_commands.py
```diff
@@ -0,0 +1,128 @@
...
+    def wait_avail_data(self):
+        initial_time = time.time()
+        curr_avail_reps = self.get_curr_avail_reps()
+        while curr_avail_reps == 0:
+            curr_avail_reps = self.get_curr_avail_reps()
+            current_time = time.time() - initial_time
+            if current_time > 10:
+                time_out = True
+                return time_out
+            time.sleep(self.wait_time_interval)
```
TobiasSpohn, line 126: This might also lock the thread for a significant amount of time, making it unresponsive. If wait_time_interval is too long this might cause problems with stopping the acquisition.

TobiasSpohn (commented):
Hi Takuya,

Great code. I added some small suggestions in the comments. 
In my comments about the while-loops, making the thread unresponsive: It might be okay to leave it as is, if the time spent in the loop is very short, so the thread is not blocked for too long.

src/qudi/hardware/fast_adc/spectrum/si_commands/buffer_commands.py
Neverhorst: Since this package is heavily used, it would also be viable to import the package as a whole without "*" in order to avoid cluttering the module namespace, i.e. `import pyspcm`

Neverhorst:
Where does the `pyspcm` dependency come from? It is not listed in `setup.py`. 
Is this derivative work of [this project](https://github.com/vuthalab/spectrum-awg/tree/master)? If so, this needs proper attribution and copyright/license disclaimers.

takuya-ulm:
> Where does the `pyspcm` dependency come from? It is not listed in `setup.py`. Is this derivative work of [this project](https://github.com/vuthalab/spectrum-awg/tree/master)? If so, this needs proper attribution and copyright/license disclaimers.

Spectrum instrumentation originally distributed some python codes with an USB stick. Recently, they shared the codes on github with MIT license:
[https://github.com/SpectrumInstrumentation/spcm](url)
Now I added this package in setup.py and modified the imports.

src/qudi/hardware/fast_adc/spectrum/si_main.py
takuya-ulm: It would be great if this error state is used when the error from the hardware is caught. I added this as TODO.

src/qudi/hardware/fast_adc/spectrum/si_commands/si_commands.py
takuya-ulm: I removed this wait function from the measurement loop(si_commander).
Now, instead of waiting in another while loop(wait_new_trig_reps), it returns in the measurement loop.

``` 
def command_process(self):
        """
        Command the main process dependent on the repetitions of unprocessed data.
        """

        trig_reps = self.cmd.process.get_trig_reps()
        processed_reps = self.processor.data.avg.num_rep
        unprocessed_reps = trig_reps - processed_reps
        if self._reps != 0 and processed_reps >= self._reps:
            return
```

takuya-ulm (commented):
@TobiasSpohn 

Thank you for your reviewing. I modified the codes and replied to your comments.
Can you have a look on them?

src/qudi/gui/time_series/time_series_gui.py
```diff
@@ -326,7 +326,7 @@ def update_channel_settings(self, enabled, averaged):
                                                     units=different_units[0])
                 self._mw.trace_plot_widget.setLabel('right',
                                                     self._channels_per_axis[1][0],
-                                                    units=different_units[1])
+                                rev                    units=different_units[1])
```
TobiasSpohn, line 329: Is this "rev" supposed to be here?

TobiasSpohn (commented):
Thanks for the modified code, it looks good from my side. I only have the one comment in time_series_gui.py, which I think might be a syntax error.

src/qudi/hardware/fast_adc/spectrum/si_main.py
```diff
@@ -0,0 +1,418 @@
...
+        If the counter is GATED it will return a tuple (2D-numpy-array, info_dict) with
+            returnarray[gate_index, timebin_index]
+
+        info_dict is a dictionary with keys :
+            - 'elapsed_sweeps' : the elapsed number of sweeps
+            - 'elapsed_time' : the elapsed time in seconds
+
+        If the hardware does not support these features, the values should be None
+        """
+        with self._threadlock:
```
timoML, line 346: This thread lock locks sketchy. It is not taken when writing to the same data in DataProcess.

PR #132, opened 2024-05-15
Introduction of a scannable laser interface

LukePiewalker:
Introduction of a scannable laser interface and dummy modules, which mimic the intended functionality.

## Description

This interface could be used to access scannable laser within a proper tool-chain for e.g. PLE measurements.
It consists of  `ScannableLaserInterface` and the `StabilizationMixin` where the Mixin could be used for scannable lasers which have some hardware capability to stabilize themselves.

In order to also incorporate, that some lasers require some working points to be set, which can be either a simple setpoint, such as a motor/grating position, or actively controlled parameters, such as a temperature or current, multiclassing with `ProcessSetpointInterface `and `ProcessControlInterface` is used in the respective dummy module.
The `ScannableLaserInterface` incorporates non-abstract properties, to check if the laser supports the respective features.

## Motivation and Context

So far scannable lasers are typically only accessed by the respective interfaces the vendor provides and having them implemented in qudi could enable additional speed up and automization of measurements.

## How Has This Been Tested?

No real test but Dummy modules have been introduced.

## Types of changes

- [ ] Bug fix
- [x] New feature

## Checklist:

### DIFF
--- a/src/qudi/interface/scannable_laser_interface.py
+++ b/src/qudi/interface/scannable_laser_interface.py
@@ -1,0 +1,164 @@
+# -*- coding: utf-8 -*-
+"""
+Interface file for lasers whose frequency can be scanned
+
+Copyright (c) 2021, the qudi developers. See the AUTHORS.md file at the top-level directory of this
+distribution and on <https://github.com/Ulm-IQO/qudi-iqo-modules/>
+
+This file is part of qudi.
+
+Qudi is free software: you can redistribute it and/or modify it under the terms of
+the GNU Lesser General Public License as published by the Free Software Foundation,
+either version 3 of the License, or (at your option) any later version.
+
+Qudi is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+See the GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License along with qudi.
+If not, see <https://www.gnu.org/licenses/>.
+"""
+
+__all__ = ('ScannableLaserInterface', 'ScannableLaserConstraints', 'LaserScanMode',
+           'LaserScanDirection', 'ScannableLaserSettings', 'StabilizationMixin')
+
+from abc import abstractmethod
+from dataclasses import dataclass
+from enum import Enum
+from typing import Union, Iterable, Optional, Tuple
+
+from qudi.core.module import Base
+from qudi.util.constraints import ScalarConstraint
+
+from qudi.interface.process_control_interface import ProcessControlConstraints, ProcessSetpointInterface, \
+    ProcessControlInterface
+
+
+class LaserScanMode(Enum):
+    CONTINUOUS = 0
+    FIXED_REPETITIONS = 1  # Includes single sweep
+
+
+class LaserScanDirection(Enum):
+    UP = 0
+    DOWN = 1
+
+
+@dataclass(frozen=True)
+class ScannableLaserConstraints:
+
+    scan_limits: Tuple[float, float]
+    scan_unit: str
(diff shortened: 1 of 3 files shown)


LukePiewalker:
Dear all 
@qku, @ksenkalla, @Nick-Grimm, @Neverhorst

I drafted up something now for the laser scanning interface and wrote (very minimal) dummy modules.
Open for comments and suggestions or criticism.

qku:
Really cool @LukePiewalker ! I like the idea of having the stabilization as a mixin. It should allow us to cover a large variety of lasers.

src/qudi/interface/scannable_laser_interface.py
```diff
@@ -0,0 +1,164 @@
...
+
+class LaserScanDirection(Enum):
+    UP = 0
+    DOWN = 1
+
+
+@dataclass(frozen=True)
+class ScannableLaserConstraints:
+
+    scan_limits: Tuple[float, float]
```
qku, line 50: Why not use also `ScalarConstraint` here? This would make things more consistent and it comes with built-in checker methods.

src/qudi/interface/scannable_laser_interface.py
```diff
@@ -0,0 +1,164 @@
...
+        pass
+
+
+@dataclass(frozen=True)
+class ScannableLaserSettings:
+
+    scan_range: Tuple[float, float]
+    scan_speed: Union[int, float]
+    scan_mode: Union[LaserScanMode, int]
+    initial_scan_direction: Union[LaserScanDirection, int]
```
qku, line 66: Personally I can't think of a situation where this is useful. May just be me though.

src/qudi/interface/scannable_laser_interface.py
```diff
@@ -0,0 +1,164 @@
...
+    scan_speed: ScalarConstraint
+    scan_modes: Iterable[Union[LaserScanMode, int]]
+
+    def __post_init__(self) -> None:
+        # TODO implement sanity checks
+        pass
+
+
+@dataclass(frozen=True)
+class ScannableLaserSettings:
```
qku, line 61: I don't think that introducing a class for the scan settings is of advantage here. Other than in the scanning probe toolchain, we don't need to save the settings together with data, right? I would go with keyword arguments for the various parameters instead.

src/qudi/interface/scannable_laser_interface.py
LukePiewalker: Was originally intending to do so, but then got a bit confused on using two scalar constraints for such a thing with upper and lower limits.

Just went for two floats out of convenience, because I thought its easier for such a simple concept.

src/qudi/interface/scannable_laser_interface.py
LukePiewalker: Are you referring to the `initial_scan_direction` part only?
One could use that to change the direction during a "running scan" and therefore overscan some sub part on the fly without needing to change the scan range.
Used it that way in the lab some time already, therefore thought might be helpful.

src/qudi/interface/scannable_laser_interface.py
LukePiewalker: Was just a fan of having things consistent. Then also one could possibly include some things such as type checking, range clipping etc in the settings class already.

src/qudi/interface/scannable_laser_interface.py
qku: We would only need a single scalar constraint here though, right? I understand the scan limits as the minimum and maximum value that the scan "position" in terms of scan units can reach. The constraint then includes the two floats for the upper and lower bound.

src/qudi/interface/scannable_laser_interface.py
qku: I see, I guess this is mostly relevant for very slow scans?

PR #122, opened 2024-02-08
add support for devices with voltage-high sensitive triggers

JohannesWes:
add support for devices with voltage-high sensitive triggers

## Description
Adds option for a config flag invert_trigger_polarity, that if checked, changes the idle_state of the NI output channels to HIGH instead of LOW and changes the edge types to FALLING instead of RISING edges

## Motivation and Context
Some devices, e.g. the microwave generators from the company Windfreak Technologies are LOW-sensitive, so trigger pulses have to consider that. A LOW idle state causes unwanted triggering. Also the edges have to be changed.
#105 

## How Has This Been Tested?
The fix has been tested by checking sweeps of the microwave signal with qudi's odmr module, using a Windfreak SynthNV Pro connected to an oscilloscope to check the output. With the fix, no unwanted triggering occurs anymore.

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change (Causes existing functionality to not work as expected)

## Checklist:

- [x] My code follows the code style of this project.
- [x] I have documented my changes in `/docs/changelog.md`.
- [ ] My change requires additional/updated documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added/updated the config example for any module docstrings as necessary.
- [x] I have checked that the change does not contain obvious errors
(syntax, indentation, mutable default values, etc.).
- [x] I have tested my changes using 'Load all modules' on the default dummy configuration.
- [ ] All changed Jupyter notebooks have been stripped of their output cells.

PR #134, opened 2024-06-06
Windfreak synthHD documentation for falling trigger edges

NicolasStaudenmaier:
## Description

Added documentation that the MW source is low active and needs a falling edge when using triggers. Using the `ni_x_series_finite_sampling_input` one can change the config option for the trigger edge. 

## Motivation and Context

Users have run into this problem. First the option to change the trigger edge has beed added with PR #122. After that users still ran into the same problem because the necessity to change the trigger edge is not stated in the hardware file. 

## How Has This Been Tested?

Not tested, only change in the docstring. 

## Types of changes

- [ ] Bug fix
- [ ] New feature

## Checklist:


takuya-ulm (approved):
Looks good to me.

PR #135, opened 2024-06-13
Fixes issue with using same pfi channel on two different ni cards for NI instreamer

prithviulm:
## Description
If there are two NI cards connected and each has an analog input on the same pfi channel, then both counter GUIs cannot be run simultaneously because the task names are identical.

## Motivation and Context
For experiments using multiple NI cards, it is possible that the same pfi channel is used and it causes the new task to have the same name as the old task. This crashes the time series gui.

## How Has This Been Tested?
Tested scanner and time series gui on my two-foci confocal setup and it worked fine. Have not tested any other modules since I don't use anything else.
Also the old core did not have this problem because there was a check to make sure task names were unique: https://github.com/Ulm-IQO/qudi/blob/ac37d66a1a89afaef455bf67699fcc189c1b60b2/hardware/ni_x_series_in_streamer.py#L874-L882

## Types of changes
- [x] Bug fix
- [ ] New feature
- [x] Breaking change (possibly breaking, I have no idea if something else depends on the existing naming convention)

## Checklist:

### DIFF
--- a/src/qudi/hardware/ni_x_series/ni_x_series_in_streamer.py
+++ b/src/qudi/hardware/ni_x_series/ni_x_series_in_streamer.py
@@ -640,7 +640,7 @@ class NIXSeriesInStreamer(DataInStreamInterface):
             # Set up digital counting tasks
             for i, chnl in enumerate(digital_channels):
                 chnl_name = f'/{self._device_name}/{chnl}'
-                task_name = f'PeriodCounter_{chnl}'
+                task_name = f'PeriodCounter_{self._device_name}_{chnl}'
                 # Try to find available counter
                 for ctr in self.__all_counters:
                     ctr_name = f'/{self._device_name}/{ctr}'


timoML:
There are similar task setups at in:
- ni_x_series_finite_sampling_io, with an id added based on the module instance. (I guess this should be the safest way)
- ni_x_series_finite_sampling_input, where it's not handled.

While we're at it: Should we decide on one mechanism to uniquely name the tasks and use this at all of these occurances?

Neverhorst:
> There are similar task setups in:
> 
> * ni_x_series_finite_sampling_io, with an id added based on the module instance. (I guess this should be the safest way)
> * ni_x_series_finite_sampling_input, where it's not handled.
> 
> While we're at it: Should we decide on one mechanism to uniquely name the tasks and use this at all of these occurances?

This again comes down to the already discussed improvement of writing an "importable NI card utility module" that every hardware implementation can use. There you could centralize commonly used code for the NI cards (like reserving resources and naming tasks).

src/qudi/hardware/ni_x_series/ni_x_series_in_streamer.py
```diff
@@ -640,7 +640,7 @@ def _init_digital_tasks(self):
             # Set up digital counting tasks
             for i, chnl in enumerate(digital_channels):
                 chnl_name = f'/{self._device_name}/{chnl}'
-                task_name = f'PeriodCounter_{chnl}'
+                task_name = f'PeriodCounter_{self._device_name}_{chnl}'
```
LukePiewalker, line 643: Looks fine as a quick fix.
I guess this could still lead to issues in case there are multiple instances of an `NIXSeriesInStreamer` using the same pfi channel. On the hardware side this should still be allowed when the task is using another clock.

Maybe one could extend it like [here](https://github.com/Ulm-IQO/qudi-iqo-modules/blob/4d6795438a2ed844956cf8ecabebfbe046cc5c2e/src/qudi/hardware/ni_x_series/ni_x_series_in_streamer.py#L573) with `id(self)` or use `self.module_uuid` as well such that the task name is unique for each instance?

NicolasStaudenmaier (approved):
As a hotfix the suggested solution works. 
We keep in mind the points of the discussion and in the future will add a wrapper for the NI functions that will be important and guarantees unique task definitions.

timoML:
This solution will break when using any functionaliy of ni_x_series_finite_sampling_input with two ni cards.

PR #136, opened 2024-06-13
Save ROI image as png

simonschmitt2:
Added the possibility to create a png file of the current ROI in the POI manager

## Description

In the POI manager, when the ‘Save ROI’ button is toggled, an additional PNG file can be created, showing the ROI image (including all the POIs). To activate this functionality, there is now a configuration option for the POI manager logic (save_roi_image_as_png: True|False). The default value for this option is False, so there should be no changes for those who do not specifically add the option. The image itself should closely resemble the plot widget in the POI manager.

## Motivation and Context

There hasn’t been any such functionality, and the workaround of taking a screenshot requires extra steps and doesn’t produce high-quality images.

## How Has This Been Tested?

Only tested with dummy config, which should not matter.

Example file: 

20240613-1501-30_roi_20240607_0707_28_126769_scan_image

## Types of changes

- [ ] Bug fix
- [x] New feature

## Checklist:


timoML:
I think saving ROI images is definetly highly needed.

To avoid code duplication, I would suggest to make use of draw_2d_scan_figure() in the scanning_data_logic:
- poimanagerlogic.set_scan_image() should then not only store the image, but keep a copy of the whole ScanData object.
- the scan axes are stored with ScanData and can be easily inserted to the plot 
- Some rework of scanning_data_logic::draw_2d_scan_figure() will be required to make it compatible with poimanagerlogic.

Minor things:
- I think default behavior should be to save the png. From my perspective, we may eben drop the ConfigOption and save it (similar to scanning_probe toolchain)
- Currently, if you don't have a confocal image, but marked some POIs, the png won't be drawn.
- The scanning_probe toolchain stores the image with the colorbar set in the GUI. It'd be nice if this was working for the poimanager, but might be a bit complicated.

simonschmitt2:
Thank you very much for your feedback!

I refactored the code so that the draw_2d_scan_figure() method in scanning_data_logic is reused. Unfortunately, it was not straightforward since there are a few differences in both save methods, such as the Points of Interest (POIs). As a result, the final code turned out a bit more complicated.

Regarding the remaining comments:

    I am against the idea that the RegionOfInterest class should hold ScanData because it contains more data than relevant, especially multiple images when multiple channels are configured. Additionally, the channel would need to be held. Furthermore, this change would require refactoring most of the poi_manager_logic file. However, it is true that, at the moment, the metadata of the image is missing in the PoiManager. As a quick workaround, I added the ScanImageMeta data class, which allows specifying this metadata. However, it is ignored when saving or loading ROIs, so a proper implementation is still to be done.
    With the data class above, the correct meta information should now be in the saved PNGs (the GUI is not updated yet, another to-do).
    As mentioned above, I did the refactoring.

Regarding the 'minor things':

    I dropped the ConfigOption.
    I am against saving a PNG when there is no image; it seems counter-intuitive.
    As far as I see, the color bar is handled only in the GUI at the moment. Refactoring this would go beyond the scope of the PR.

timoML:
During final testing, I encountered an issue when coming from StatusVars of current main:
1. Checkout main
2. Delete all StatusVar from poi_manager_logic
3. Create a new scan in scanning_probe
4. Start poi_manager_gui and hit "Get confocal image"
5. Close Qudi
6. Checkout this PR

I think expected behavior in poi_manager_logic should be to start over with a fresh poi, if the StatusVars can't be parsed.

```
 Error while settings status variables:

TypeError: Parameter to generate ScanImageMeta instance from must be of type dict.

  File "C:\Users\Timo\AppData\Local\conda\conda\envs\qudi-newrepo\lib\site-packages\qudi\core\module.py", line 338, in _load_status_variables
    value = var.constructor_function(self, value)

  File "C:\Software\qudi\qudi-iqo-modules\src\qudi\logic\poi_manager_logic.py", line 1308, in dict_to_roi
    return RegionOfInterest.from_dict(roi_dict)

  File "C:\Software\qudi\qudi-iqo-modules\src\qudi\logic\poi_manager_logic.py", line 321, in from_dict
    scan_image_meta=ScanImageMeta.from_dict(dict_repr.get('scan_image_meta')),

  File "C:\Software\qudi\qudi-iqo-modules\src\qudi\logic\poi_manager_logic.py", line 347, in from_dict
    raise TypeError(f'Parameter to generate {ScanImageMeta.__name__} instance from must be of type '
```

simon1schmitt:
Hi timoML,

Thank you very much for your review and for identifying the issue. However, I am not entirely sure how to tackle it. As far as I understand, the` from_dict` method is used to create the ROI-StatusVar. First, I would expect this method to throw an exception if the wrong data type is input (and not just create a new ROI). Second, this method does not seem to be solely for creating the StatusVar, so I would expect it to also work when called from the manager, for example. In this case, I would expect to see an exception and not just a new ROI being created as well.

Currently, I think that the `_load_status_variables` in `\src\qudi\core\module.py` should be responsible for handling failed attempts to create StatusVars. Is there perhaps some agreement on how to handle these cases? Or is there another module where I could look to see how the handling is done there?

timoML:
I think the StatusVar constructor is the best place to handle this (see my commit).

timoML (approved):
Looks good, thank you.

PR #138, opened 2024-06-18
Remove old odmr fits

simonschmitt2:
Remove old fits when starting a new ODMR measurement

## Description

As discussed in https://github.com/Ulm-IQO/qudi-iqo-modules/discussions/137, all fits are removed when starting a new ODMR measurement.

## Motivation and Context

Old fits were not removed when starting a new measurement, which could be annoying especially when the frequency range changed between the old and the new measurement.

## How Has This Been Tested?

Tested with dummy configuration, which should not matter

## Types of changes

- [x] Bug fix
- [ ] New feature

## Checklist:

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -12,6 +12,9 @@
 - Improved support for Stanford Research Systems signal generators
 - Expanded documentation of the microwave interface
 
+### Bugfixes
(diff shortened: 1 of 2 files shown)


Neverhorst (approved):
Looks good to me. Thank you.

PR #139, opened 2024-06-18
POI manager auto-find fix

simonschmitt2:
POI manager auto-find fix

## Description

When running the (cool) auto-find feature in the POI manager, the POIs did not match the bright spots but appeared to be rotated by 90°. The issue seemed to stem from an unnecessary transpose operation on the image data, which has now been removed. Additionally, if the POI diameter was smaller than the image step size, an exception was thrown due to a filter size of 0. This has been resolved by ensuring that the filter size is at least 1.

## Motivation and Context

After these changes the feature seems to run as expected

## How Has This Been Tested?

Tested only with dummy configuration, which should not matter

## Types of changes

- [x] Bug fix
- [ ] New feature

## Checklist:

### DIFF
--- a/src/qudi/logic/poi_manager_logic.py
+++ b/src/qudi/logic/poi_manager_logic.py
@@ -1265,7 +1265,7 @@ class PoiManagerLogic(LogicBase):
 
     def _local_max(self, scan):
         scan = np.asarray(scan, order="C")  # scan has to be a 2-D array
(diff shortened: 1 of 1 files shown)


NicolasStaudenmaier (approved):
Thank you Simon for finding this. Looks good and the auto-find tool works well with real data.

PR #142, opened 2024-07-10
Fastcomtec get sweeps

NicolasStaudenmaier:
Get the current number of sweeps and runtime of the FastComtec MCS6 while running a measurement. 

## Description

The FastComtec counter model MCS6 has the possibility to read the number of sweeps and runtime when a measurement is performed. This has not been implemented so far and is added with this PR. 

## Motivation and Context

The info_dict when getting the data trace returns `None`. Also when the measurement is saved the information about the elapsed sweeps will not be saved. By simply querying the sweeps and runtime of the fastcounter the values are available for the user especially once the measurement is saved. 

## How Has This Been Tested?

Tested on my setup with real hardware MCS6A. 

## Types of changes

- [ ] Bug fix
- [x] New feature

## Checklist:

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -8,6 +8,7 @@
 - "NFiniteSamplingInput supporting both trigger polarities via ConfigOption
 
 ### New Features
+- Get current sweeps and runtime of fastcomtec fastcounter MCS6 when pulling the data trace. 
 - Re-introduced tilt correction (from old core) to the scanning probe toolchain.
 - Improved support for Stanford Research Systems signal generators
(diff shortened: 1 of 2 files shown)


rmgonzal:
I tested with my setup FastComtec MCS6 and it works (saved measurement with `info_dict = {'elapsed_sweeps', 'elapsed_time'}`) also the values are available in the `PulsedMeasurementGui` rather than -1.
scree
@timoML Could you check this with your hardware?

rmgonzal (approved):

PR #145, opened 2024-07-30
Fixed some bugs in laserquantum_laser hardware file

TobiasSpohn:
## Description

- When loading the laser toolchain with the laserquantum laser it will fail upon activation. I fixed a typo when querying the shutter state `get_shutter_state` . 
- I removed the `PSUTypes.FPU` in `get_extra_info` as it is not present in `PSUTypes` and I could not find a matching power supply on laserquantums website
- Fixed the return value of `get_power_setpoint`. Upon setting the power a variable `_last_set_power` will be set that stores the last setpoint. When querying the power setpoint this will be returned instead of the current power, which made the setpoint spinbox in the gui unusable. Upon module activation the currently set power will be retrieved. This will be close to 0 W if the laser is off or it will be the currently set output power.

## Motivation and Context
- First two changes prohibit the use of the laser toolchain with the `laserquantum_laser` hardware file
- The command that was previously sent to the hardware within the if statement that I removed is not documented in the manuals for the different PSU types and returns an Error if sent to the device. Thus I removed it as no information can be gained by this query.
- Setting the currently displayed setpoint to the current output power will result in an unusable setpoint spinbox, which constantly fluctuates in value.

## How Has This Been Tested?
I tested it on my setup using a laserquantum gem laser with SMD12 PSU.

## Types of changes
- [x] Bug fix
- [ ] New feature
- [x] Breaking change (Causes existing functionality to not work as expected) (If anybody relied on the extra information output)

## Checklist:

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -6,6 +6,9 @@
 
 ### Bugfixes
 - "NFiniteSamplingInput supporting both trigger polarities via ConfigOption
+- Fixed typo in `LaserQuantumLaser` when returning the `ShutterState`
+- Removed `PSUTypes.FPU` if statement in `LaserQuantumLaser` as it is not an existing `PSUTypes`
+- Fixed `LaserQuantumLaser.get_power_setpoint` to return last set power instead of current power
 
 ### New Features
 - Re-introduced tilt correction (from old core) to the scanning probe toolchain.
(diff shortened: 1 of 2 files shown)


TobiasSpohn:
The issues are solved in PR #16.

PR #148, opened 2024-08-06
Scanner customnametag

prithviulm:
## Description
Added a UI element in the scanner GUI for the user to specify a custom save file name for saved scans. Note that this is only appended to the default nametag that is generated which contains the timestamp.

## Motivation and Context
Some users do multiple confocal scans and it is difficult to keep track of the scans since they are only saved with timestamps in the file name.

## How Has This Been Tested?
Tested on Ubuntu 22.04.4LTS with dummy modules and on Windows 10 (custom config for experiments) and it works fine in both. Note that the UI changes made here were based on those in the ODMR GUI.

## Types of changes

- [ ] Bug fix
- [x] New feature

## Checklist:

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -11,6 +11,7 @@
 - Re-introduced tilt correction (from old core) to the scanning probe toolchain.
 - Improved support for Stanford Research Systems signal generators
 - Expanded documentation of the microwave interface
+- Added option to specify custom save file name in scanning GUI
 
 ### Bugfixes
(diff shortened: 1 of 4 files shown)


timoML (approved):
Looks good. Can you document the maintenance team discussion on a long term saving overhaul somewhere (e.g. Issues)?

prithviulm:
> Looks good. Can you document the maintenance team discussion on a long term saving overhaul somewhere (e.g. Issues)?

This is now tracked under #150.

PR #149, opened 2024-08-09
bugfix on unsaved fit figure in QDPlotter

takuya-ulm:
## Description

The plot in qdplot was not saved when fitting was used 

This was caused by improper handling of fit_results.
In qdplot_logic.py, _last_fit_results is defined under QDPlotFitContainer as a dictionary.
When _plot_figure of QDPlotLogic is called for saving, this fir_results is used to retrieve the result.high_res_best_fit. 
Since the instance of ModelResult from lmfit is stored in the values of that dictionary, values() should be used.

## How Has This Been Tested?

Firstly, data is set by set_data method. After fitting with a function on gui, save_data method was called.

## Types of changes

- [x] Bug fix
- [ ] New feature

## Checklist:

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -33,6 +33,7 @@ Configuration for time series toolchain needs changes as well. See `default.cfg`
 docstrings.
 
 ### Bugfixes
+- Fix failure of saving plots in `QDPlotLogic` when fiting is used.
 - Basic data saving in `TimeSeriesReaderLogic` works now.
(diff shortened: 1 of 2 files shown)


NicolasStaudenmaier (approved):
Tested and approved. Thanks Takuya.

timoML:
Side note: In pulsed we are saving the fit result in the metadata. In qdplotlogic, the fit result is currently only put onto the plot.

Issue #154, opened 2024-09-05
[Bug] Can't load POI Manager toolchain


TobiasSpohn:
### Version

current main

### What is affected by the bug?

Since #136 one can no longer start the POI manager toolchain. `poi_manager_logic` will throw an exception during `on_activate`. If no ROI has been saved as StatusVar before the module will not activate. The default creation of the ROI does not seem to work. I have inserted a warning message at the beginning of on_activate and as the log shows `self._roi` is `None`. By switching to one commit before the merge of PR #136, not clearing the StatusVars and then activating the module on the current commit, POI toolchain launches.

### When does the bug occur?

It occurs during activation of `poi_manager_logic` with all StatusVars cleared.

### How do we replicate the issue?

- Clear StatusVars of `poi_manager_logic`
- activate `poi_manager_logic`

### Expected behavior

`poi_manager_logic` starts and the correct default ROI is created if none is already present.

### Relevant log output

```shell
2024-09-05 17:05:48 info qudi.core.modulemanager Activating logic module "poi_manager_logic.PoiManagerLogic"
2024-09-05 17:05:48 info qudi.core.modulemanager Activating logic module "scanning_probe_logic.ScanningProbeLogic"
2024-09-05 17:05:48 info qudi.core.modulemanager Activating hardware module "dummy.scanning_probe_dummy.ScanningProbeDummy"
2024-09-05 17:05:48 debug qudi.core.threadmanager Creating thread: "mod-hardware-scanner_dummy".
2024-09-05 17:05:48 debug qudi.core.threadmanager Creating thread: "mod-logic-scanning_probe_logic".
2024-09-05 17:05:48 info qudi.core.modulemanager Activating logic module "scanning_optimize_logic.ScanningOptimizeLogic"
2024-09-05 17:05:48 debug qudi.core.threadmanager Creating thread: "mod-logic-scanning_optimize_logic".
2024-09-05 17:05:48 info qudi.core.modulemanager Activating logic module "scanning_data_logic.ScanningDataLogic"
2024-09-05 17:05:48 debug qudi.core.threadmanager Creating thread: "mod-logic-scanning_data_logic".
2024-09-05 17:05:48 debug qudi.core.threadmanager Creating thread: "mod-logic-poi_manager_logic".
2024-09-05 17:05:48 warning qudi.logic.poi_manager_logic.PoiManagerLogic self._roi=None
[...]

PR #155, opened 2024-09-05
Fix for optimization sequence in scanning toolchain

TobiasSpohn:
## Description
Fixes that the optimizer could not perform optimization sequences other than the default [2, 1], e.g. ["xy","z"]. By changing the checker function of the ConfigOption `ScannerGui.optimizer_plot_dimensions` to a constructor that correctly checks the constraints. Furthermore, the selection of available sequences is now handled by `scanning_optimize_logic`, the GUI only access it.
A signal is added to lay the ground for future implementation of adjusting the optimizer sequence during runtime.

## Motivation and Context
The `ScannerGui.optimizer_plot_dimensions` ConfigOption is there to adjust the optimization sequence to an arbitrary combination of 2D and 1D optimizations within the constraints of the hardware (i.e. number of axes). It should be possible with e.g. 3 Axes to perform the default 2D and 1D optimization, but also any subset of these. E.g. only a 1D or multiple 1D optimizations. Due to a bug in the checker method of the ConfigOption it was not possible to select only 1D optimizations. Furthermore, it was not possible to change `ScanningOptimizeLogic._scan_sequence` once it was stored as a StatusVar and the default fallback for the StatusVar was hardcoded to a sequence of [2,1].

## How Has This Been Tested?
Default config file

## Types of changes
- [x] Bug fix
- [ ] New feature
- [x] Breaking change (Causes existing functionality to not work as expected) -> The ConfigOption moved to the logic

## Checklist:
- [x] I have tested my changes using 'Load all modules' on the default dummy configuration. -> POI manager can't load due to #154 


TobiasSpohn:
With Hotfix #156 load all modules completes without errors.

docs/changelog.md
```diff
@@ -7,6 +7,7 @@
   implementing this interface.
 - Add back scan data and back scan settings getter and setter methods
   to `qudi.interface.scanning_probe_interface`.
+- Moved ConfigOption ScannerGui.optimizer_plot_dimensions to qudi.logic.scanning_optimize_logic and optimizer sequence creation is solely handled by logic
```
timoML, line 10: need to mention the new name of the ConfigOption "optimizer_sequence_dimensions"

timoML:
That's definitely an improvement.
While testing I encountered that, currently, we get a lot of StatusVar related errors when configuring a 2-axis scanner_dummy.
Maybe we should put in some work into fixing these.
After StatusVar clearing, the behavior with 2-axis scanners is clearer now.

TobiasSpohn:
I rewrote the image generation of the dummy. The dummy can now have an arbitrary number of axes.

src/qudi/hardware/dummy/scanning_probe_dummy.py
```diff
@@ -583,24 +761,46 @@ def __stop_timer(self):
...
+
+        return scan_vectors
+
+    @_spot_density.constructor
+    def spot_density_constructor(self, spot_density: float) -> float:
+        volume_edges = [abs(pos_range[1] - pos_range[0]) for pos_range in self._position_ranges.values()]
+        volume = 1
+        for edge in volume_edges:
+            volume *= edge
+        if volume * spot_density ** len(self._position_ranges.keys()) > 80e3:
```
timoML, line 793: Maximum spot number should be a hardware constraint.

TobiasSpohn: I implemented it as a constraint and therefore subclassed the `ScanConstraint` class. This however results in the maximum spot number check to occur only during `on_activate` of the module and not during the constructor of the `ConfigOption`.

src/qudi/logic/scanning_optimize_logic.py
```diff
@@ -389,3 +446,19 @@ def _set_default_scan_settings(self):
         self._scan_frequency = {ax.name: max(ax.frequency.minimum, ax.frequency.maximum / 100) for ax in axes.values()}
         self._back_scan_resolution = {}
         self._back_scan_frequency = {}
+
+    def _set_scan_sequence(self):
```
timoML, line 450: Ambigous name of function; it's not a setter. Maybe _set_default_scan_sequence()?

src/qudi/gui/scanning/scannergui.py
```diff
@@ -439,6 +418,22 @@ def _init_static_dockwidgets(self):
         self.tilt_correction_dockwidget.visibilityChanged.connect(self._mw.action_view_tilt_correction.setChecked)
         self._mw.action_view_tilt_correction.triggered[bool].connect(self.tilt_correction_dockwidget.setVisible)
 
+    def _set_optimizer_dockwidget(self):
```
timoML, line 421: Ambigous name of function; it's not a setter. Maybe _init_optimizer_dockwidget()?

timoML:
Except for minor notes above, it looks really good now. We still should:
- run an overnight test with continuous optimization to test the changed BlockingQueuedConnection
- maybe run ruff over the files (@prithviulm).

prithviulm:
I made a branch here https://github.com/Ulm-IQO/qudi-core/tree/embracing_ruff for developing the ruff config we intend to use. @TobiasSpohn could you try it out? The config is within pyproject.toml. Installation instructions are in the commit message, see here for quick reference https://github.com/Ulm-IQO/qudi-core/commit/7a38bc10518604a7fc6b29ffe5c2a9d37ff421b9

TobiasSpohn:
As discussed in the software meeting, all debug and `Qt.BlockingQueuedConnection` changes have been reverted and the branch is ready for testing with the fix for scanning. The thread deadlock issue will need some further investigation and some other PR to fix.

timoML (dismissed):
Looks good now, thank you.

src/qudi/gui/scanning/scannergui.py
TobiasSpohn: changed the name

src/qudi/logic/scanning_optimize_logic.py
TobiasSpohn: changed the function name

timoML (approved):
Thank you.

PR #156, opened 2024-09-06
Hotfix StatusVar Poimanager

timoML:
Fixes an issue that prevents loading of the module with default StatusVar.


TobiasSpohn (approved):
Fixes the issue, thanks.

PR #161, opened 2024-10-02
Fixed counting_length when in gated mode for T1 sequencing predefined method

TobiasSpohn:
## Description
Fixes the `counting_length` of the `measurement_information` dict when using a gated fastcounter and the `t1_sequencing `predefined method to be the length of one readout.

## Motivation and Context
The `count_length` is previously calculated inside the sequence step generation loop. This just sums up all `PulseBlockEnsemble` lengths yielding a `count_length` equal to the length of the entire pulse sequence. For gated mode a `count_length` of only one readout is required to configure the fastcounter.

## How Has This Been Tested?
On a confocal setup with a FastComtec MCS6A fastcounter.

## Types of changes
- [x] Bug fix
- [ ] New feature

## Checklist:


timoML (approved):
Looks good, thank you.
Long term, we should generalize this behavior in a future `self._get_sequence_count_length()`

PR #162, opened 2024-10-03
Keysight AWG update

TobiasSpohn:
## Description

- Enables DMC output mode for Keysight M8195A AWG
- Fixes bugs in sequence mode so it is usable for Keysight M8195A AWG by removing unneeded / deprecated constraint `sequence_order`
- Fixes setting of digital amplitude on activation of the module

## Motivation and Context

- With the current Keysight M819A module it is not possible to use DMC output mode, which @timoML added.
- It was not possible to upload sequences to the M8195A as a `sequence_order` constraint was not implemented.
- Upon activation digital channel amplitude and offset are set. A ConfigOption specified to which levels these were set, however there was a bug in the calculation of the amplitude to set.

## How Has This Been Tested?
Confocal setup with a Keysight M8195A AWG

## Types of changes
- [x] Bug fix
- [x] New feature

## Checklist:


timoML:
I tested on setup 3 with a scope. Looks fine, especially trigger levels behave and warn as expected.

timoML (dismissed):
Looks good now, thanks.

PR #165, opened 2024-10-28
fix checker for optimizer plot dimensions

qku:
## Description

fix a bug in the checker of the `optimizer_plot_dims` config option of the scanner gui: allow for only 2D optimization (no 1D optimization)

## Motivation and Context

I noticed this bug while setting up confocal for a system with only X and Y control, and only manual/external Z control. Setting the config option to `[2, 2]` did not work.

## How Has This Been Tested?

tested on an NI card system, but should be possible to test with scanner dummy as well

## Types of changes

- [x] Bug fix
- [ ] New feature

## Checklist:

### DIFF
--- a/src/qudi/gui/scanning/scannergui.py
+++ b/src/qudi/gui/scanning/scannergui.py
@@ -123,7 +123,7 @@ class ScannerGui(GuiBase):
     _optimizer_plot_dims: List[int] = ConfigOption(
         name='optimizer_plot_dimensions',
         default=[2, 1],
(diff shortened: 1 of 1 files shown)


timoML:
We can probably merge this small change now. There is a bigger [overhaul](https://github.com/Ulm-IQO/qudi-iqo-modules/pull/155) coming soon that'll deprecate said ConfigOption.

timoML (approved):
Thanks for finding this bug.

PR #175, opened 2024-11-29
Save and restore window position in various module GUIs

TobiasSpohn:
## Description
Adds the `GuiBase` builtin methods `_save_window_geometry` and `_restore_window_geometry` to the following GUIs:
`NVCalculatorGui`, `QDPlotterGui`, `OdmrGui`, `TimeSeriesGui` and `PulsedMeasurementGui`.
This enables these GUIs to restore the window position and size after a restart.

## Motivation and Context
When closing and opening (e.g. upon qudi restart) the main windows of `NVCalculatorGui`, `QDPlotterGui`, `OdmrGui`, `TimeSeriesGui` and `PulsedMeasurementGui` the windows spawn at a random position. This can be annoying if one has multiple windows open over multiple screens and has to quickly restart qudi, as these windows will change position again.

## How Has This Been Tested?
On a confocal setup and on dummy modules.

## Types of changes
- [ ] Bug fix
- [x] New feature

## Checklist:


NicolasStaudenmaier (approved):
The save and restore window geometry is a nice feature to improve workflow with Qudi. It is still not used in all modules. In the future we should think about making this an integrated default behavior for all modules to avoid confusing individual behavior of each module. 
For now we can merge merge this PR as I believe it improves the user experience of the modified modules.

PR #176, opened 2024-11-29
Qdplot GUI fit fix

TobiasSpohn:
## Description
When adding a new plot, existing plots will no longer close their Fit widget, displaying the fit results, when in tabbed mode. The feature is respecting the display state of each plot before the new plot was added, if a user hid the result manually it will not be shown after the addition of the new plot.

## Motivation and Context
Whenever a new plot is added, the `restore_view` function is called. This function will reset the state of showing fits on all plots to an initial `False` state. This can be annoying when adding new plots and comparing them with each other, if suddenly the fit results disappear.

## How Has This Been Tested?
Confocal setup and dummy modules.

## Types of changes
- [x] Bug fix
- [ ] New feature

## Checklist:

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -20,6 +20,7 @@
 - Fixed psu-dependent bugs for laser_quantum_laser
 - Laser logic does not automatically start query loop, gui will still start it on startup
 - Fixed `t1_sequencing` predefined method's `counting_length` for gated mode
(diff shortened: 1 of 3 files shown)


takuya-ulm (approved):
Thanks Tobi for the bug fix.
I checked the code and confirmed the proper behavior by testing linear fitting.

PR #177, opened 2024-12-09
Correctly save generation_method_parameters of PulseSequences

TobiasSpohn:
## Description
Adds `generation_method_parameters` variable to the `PulseSequence` class, so the `pulsed_measurment` toolchain can correctly save to file.

## Motivation and Context
With commit 6aa063c the `PulseBlockEnsemble.generation_method_parameters` were introduce in order to separate the method parameters from the other measurement information. This was not done for the `PulseSequence` class, resulting in `generation_method_parameters` never being updated for a `PulseSequence` in the toolchain. When saving the measurement this resulted in no `generation_method_parameters` or the ones from the previous `PulseBlockEnsemble` measurement, being saved (Depending whether there was a `PulseBlockEnsemble` since startup).

## How Has This Been Tested?
Dummy and confocal setup.

## Types of changes
- [x] Bug fix
- [ ] New feature

## Checklist:

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -21,6 +21,7 @@
 - Laser logic does not automatically start query loop, gui will still start it on startup
 - Fixed `t1_sequencing` predefined method's `counting_length` for gated mode
 - `QDPlotterGui` will continue to show fit results if new plot was added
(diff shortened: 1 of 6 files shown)


timoML:
This extends #42.
I don't like the code duplication between a PulseSquence, PulseBlockEnsemble object too much. If we don't have a more elegant idea, I think it's ok for now.

PR #178, opened 2024-12-16
Fix scan deadlock

timoML:
Solves a deadlock in the scanning toolchain.

## Description
The deadlock could occur, when scanning_optimize_logic is calling scanning_probe_logic.start_scan().
In this case, the scanning_optimize_logic thread holds a lock (scanning_probe_logic._thread_lock) when entering start_scan().
This lock can prevent __start_timer() to invoke the timer on the scanning_probe_logic via a BlockingQueuedConnection, since this threads wait on lock-release.
Rather surprisingly, the deadlock conditions seems to be timing-related and is not always found.
It might be the reason, why on some setups the optimizer freezes after hours of operation.

The solution is to call __start_timer() outside the thread lock. This is is safe, as the thread lock block before is guaranteed to have executed and __start_timer is not using any shared ressources.

## Motivation and Context
For my dummy config, this deadlock can be reproduced by a fresh start of the scanning toolchain and start an optimization without any other prior action.
The gui hangs like this, with no way to recover:
image

## How Has This Been Tested?
Dummy scanning toolchain, optimizing continuously via POIManager for > 24 h.

## Types of changes
- [x] Bug fix
- [ ] New feature

## Checklist:


timoML:
We should put a warning on similar code constructs to not call __start_timer() from within a thread lock.

timoML:
I tested your additional changes once again on setup3, looks good. It's way better to not move to all saved targets on restoring history.

TobiasSpohn (approved):
Thanks for the great fix.

I ran it on a confocal setup for 10 days now and I did no longer experience any deadlocks when starting the optimizer.

timoML:
Can you pull one more time (after merging current main into here)?

TobiasSpohn (approved):
Still works fine, thanks

TobiasSpohn (approved):
Thanks for merging main into it, still works nicely.

Issue #170, opened 2024-11-07
[Bug] config option error using a NI scanner with two AO outputs (2D scanning only)


syim70:
### Version

current main

### What is affected by the bug?

I tried to use a NI scanner with two AO outputs, but config file's scanner gui option "optimizer_plot_dimensions: [2]" does not work. It seems like Qudi requires "z axis" even if there is no additional AO.

### When does the bug occur?

1. Latest release (both qudi-core and qudi-iqo-modules):  "optimizer_plot_dimensions: [2]" is not working even with scanner_dummy.
2. Optimizer_fix_1D branch ([#155](https://github.com/Ulm-IQO/qudi-iqo-modules/tree/optimizer_fix_1d)): "optimizer_plot_dimensions: [2]" is working with scanner_dummy, but not with a real hardware (NI DAQ with two AOs).

### How do we replicate the issue?

If users choose only 2D scanning, ni scanner (or optimizer) should not require more axes.

### Expected behavior

Please, let me have an options in config file with 2D scanning only.

### Contact Details

timoML:
Can you please paste the full error that you experience with #155 and nicard?

syim70:
As you noted, I double checked the branch name "1d_scanner_optimize."
Here are two cases.
First, ni_scanner, with a full list of AOs in config file (as default).
Second, ni_scanner, without unavailable AOs (and z axis) in config file.
Thanks.

**1. For ni_scanner, with a full list of AOs**
Invalid analog source channels encountered. Following sources will be ignored:
  ao2, ao3
Valid analog input channels are:
  ai0, ai1, ai2, ai3, ai4, ai5, ai6, ai7, ai8, ai9, ai10, ai11, ai12, ai13, ai14, ai15
Exception during activation:
Traceback (most recent call last):
  File "...\lib\site-packages\qudi\core\module.py", line 301, in __activation_callback
    self.on_activate()
  File "...\qudi-iqo-modules\src\qudi\hardware\ni_x_series\ni_x_series_finite_sampling_io.py", line 246, in on_activate
    raise ValueError(
ValueError: The channels "ao2, ao3", specified in the config, were not recognized.

Traceback (most recent call last):
  File "...\lib\site-packages\qudi\core\modulemanager.py", line 201, in activate_module
    self._modules[module_name].activate()
  File "...\lib\site-packages\qudi\core\modulemanager.py", line 502, in activate
    module.activate()
[...]

PR #179, opened 2025-01-07
Hotfix for issue in #170

TobiasSpohn:
## Description
Fixes the initialization error occuring when configuring scanner with less than 3 axes as described in #170 .

## Motivation and Context
A bug caused in `_set_default_scan_sequence` to always raise the `NotImplementedError` in `_allowed_sequences` if the default `_optimizer_sequence_dimensions` were used with scanners that have less than 3 axes configured.

## How Has This Been Tested?
Dummy session having only x and y axis configured.

## Types of changes

- [x] Bug fix
- [ ] New feature

## Checklist:


timoML:
Looks good, thank you.
Can you quickly check that with StatusVar from a 3 axis scanner:
- scanning_probe_logic and scanning_gui crash on load
- the commits I added solve this

timoML:
In future, we should think about a more elegant solution of check_scan_settings in scanning_probe_logic and scanning_optimize logic. For now, this hotfix is a big improvement already.

PR #180, opened 2025-01-13
Tektronix AWG hardware file update to include the B series of AWGs

michaelOlney-Fraser:
Tektronix produces the AWG70K series. The old ones are labeled 'A', and there is a newer series labelled 'B'. The current hardware file only allows for the A series. This pull request contains changes to the tektronix_awg70k hardware file to include the B series. Importantly, the new series is basically the same from a hardware standpoint, and the only real difference is the maximum length of loaded waveforms. This is handled by querying the hardware so no changes are necessary to handle the hardware change.

## Description
In the hardware file, there are a few if statements which read:

if self.awg_model == 'AWG7000#A' :

I simply changed these if statements to be:

if self.awg_model in ['AWG7000#A', 'AWG7000#B'] :  

(where # can be either 1, for the single channel AWGs, or 2 for the 2 channel AWGs).

I also changed the function _has_sequence_mode() to handle the B series of AWGs. 

## Motivation and Context
This change is required because using the new tektronix AWGs breaks the pulsed gui.

## How Has This Been Tested?
This change is relatively minor and so has undergone minimal testing. It should be tested on a setup using the A series AWG before it is released. I have successfully loaded all modules that connect to this hardware without issue.

## Types of changes
- [ ] Bug fix
- [ x ] New feature

## Checklist:

- [ x ] My code follows the code style of this project.
- [ x ] I have documented my changes in `/docs/changelog.md`.
- [ x ] I have added/updated the config example for any module docstrings as necessary.
- [ x ] I have checked that the change does not contain obvious errors
- [ x ] I have tested my changes using 'Load all modules' on the default dummy configuration.
- [ x ] All changed Jupyter notebooks have been stripped of their output cells.


TobiasSpohn (dismissed):
Thanks for the additional hardware support, Michael.

I tested generating and running measurements in waveform and sequence mode on a confocal setup with a 70001A to make sure it runs without problems on already supported hardware.

TobiasSpohn (approved):
thanks

PR #181, opened 2025-01-17
New magnet toolchain 

timoML:
This PR introduces a new magnet toolchain, including a new MagnetInterface.

## Description
Hardware that implements the new MagnetInterface can be either a current-like controller or a motor that moves a permanent magnet. To abstract the concrete implementation away, both types just get_control() and set_control().

## How Has This Been Tested?
On setup3, with a connected Zaber 3D linear stage.

## Screenshots:
The toolchain includes a GUI (derived from the scanning_probe toolchain) that scans the control values while obtaining data from a FOM, which can be an arbitrary function (e.g. APD counts).

image

## Types of changes
- [ ] Bug fix
- [x] New feature

## Checklist:


Neverhorst:
The crosshair size can be set as percentile of the scan range.
It is an optional `__init__` parameter of [`qudi.util.widgets.plotting.plot_widget.DataSelectionMixin`](https://github.com/Ulm-IQO/qudi-core/blob/c752fd8be3c2e3c65e7c5a839273ccd736314427/src/qudi/util/widgets/plotting/plot_widget.py#L94).
Alternatively it is an existing [min_crosshair_size_fraction `ConfigOption`](https://github.com/Ulm-IQO/qudi-iqo-modules/blob/425029758a6c831f8d95e6569bfc2a1bfb64071c/src/qudi/gui/scanning/scannergui.py#L111) of the base ScannerGui class.

`set_activity_state` probably serves as electrically enabling/disabling individual motor axes but this should of course be documented.

simonschmitt2:
Thank  you for your feedback @Neverhorst,

I managed to set the size of the crosshair equal to the size of a single pixel in the scan. Furthermore, after discussing with @timoML, I changed the following:

- Renamed the window of the magnet GUI to 'Magnet'.
- Removed 'Display Settings' and 'Scan Settings' under 'Settings' in the app bar (and all the related code in MagnetGui).
- Removed 'block_while_moving' in MagnetInterface since not used in the moment. 
- Removed 'set_activity_state' in MagnetInterface since not used in the moment.
- Added a dataclass which should be returned by 'get_status' in MagnetInterface.
- Fixed incorrect method calls of constraints in MagnetInterface
- Removed mention of velocity in 'set_control' of MagnetInterface.
- Removed all unused imports in MagnetLogic.
- Removed all unused imports in MagnetGui.
- Removed all unnecessary appearances of 'optimizer' (never used -> clean up) in MagnetGui
- The crosshair in MagnetGui now fits the size of a single scan pixel
- Added a new method 'is_ready' to MotorInterface which was kind of needed in MagnetInterface (method makes sense for a motor anyway, added it to all existing Motor-Implementations, but mainly without meaningful functionality since I am not familiar with most motors)
- Added a MagnetMotorInterfuse which should work for all motors
- More refactoring and code clean-up

However, there are still quite a few 'todo' comments in the current version, and there are many areas where documentation and overall code clarity could be improved. Unfortunately, I currently do not have the capacity to address all of these issues. That said, I still believe this version is quite useful and is a significant improvement over having no tool at all.

TobiasSpohn (dismissed):
Thanks for the addition, let's add it as it is better to have some magnet control toolchain than none at all.

PR #183, opened 2025-01-31
Adding missing waiting element to `t1_sequencing` predefined method

TobiasSpohn:
## Description
When testing for #180, I discovered that `generate_t1_sequencing` is missing the `waiting_element` most of our pulse sequences have after the `laser_element` and `delay_element`.
In my particular case this caused the laser pulse to reach over into the next sweep, if only one datapoint is acquired.

t1_sequencing_missing_waiting_time

With the fix it works as expected.

t1_sequencing_fixed_waiting_time

Probably, this stems from a badly calibrated laser delay or laser safety, so could be fixed by changing these values as well. However, I thought with this fix one could bring this pulse sequence in line with the others, adding the `waiting_element`.

## How Has This Been Tested?
Same confocal setup #180 was tested with.

## Types of changes

- [x] Bug fix
- [ ] New feature

## Checklist:

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -29,6 +29,7 @@
 - Fixed `ScanningOptimizeLogic` crashing on first start when using scanner with less than the default 3 axes configured
 - Fixed Keysight M8195A AWG sequence mode
 - Fixed setting of digital channel amplitude of Keysight M819X AWG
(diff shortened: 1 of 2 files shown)


timoML (approved):
Definitely unintended, thanks for fixing.

PR #185, opened 2025-02-05
Embrace pyproject.toml and remove setup.py

prithviulm:
## Description
As done in https://github.com/Ulm-IQO/qudi-core/pull/111, removed `setup.py` and moved everything into `pyproject.toml` instead. Also updated the test workflow to use `build`.

## Motivation and Context
See https://github.com/Ulm-IQO/qudi-core/pull/111.

## How Has This Been Tested?
- Have tested this locally in a new Python 3.10.16 environment. All dummy modules load fine.
- The test release workflow needs to be tested with this.

## Types of changes
- [ ] Bug fix
- [ ] New feature

## Checklist:


prithviulm:
We can consider merging this after the next release to avoid slowing down our release pipeline in the event of unforseen issues with the build workflow.

TobiasSpohn (approved):
Thanks for updating to a `pyproject.toml` install method. I tested it with `python 3.10.16` and installation runs fine. Loading all modules also works.
The changes in `release_test_pypi.yml`, we will have to check whether they work after merging and quickly bumping the version number. However, they are in line with the `qudi-core` workflow, so should probably work.

PR #197, opened 2025-09-05
added try-except block in sample_pulse_block

mohitulm:
Enclosed the code of the sample_pulse_block_ensemble method of sequence_generator_logic.py in a try block to unlock the module state when something breaks.

## Description
When generating a sequence in the pulsed module with an invalid configuration, the module state can become locked, requiring a restart of Qudi. One of the causes is invalid parameter values in the sample_pulse_block_ensemble method. 
The introduced except block handles the exceptions like the existing one for memory error.  

## Motivation and Context
One way to reproduce the issue is to set negative values to predefined method parameters (e.g., setting 'spinlock_length' for the 'hh_pol' method and clicking 'Generate Sample and Load'). The module gets stuck at loading the sequence, and Qudi has to be restarted. 

## How Has This Been Tested?
This has been tested with a test that automatically loads all the predefined methods with negative values for numerical parameters. 

## Types of changes
- [x] Bug fix
- [ ] New feature

## Checklist:

### DIFF
--- a/src/qudi/logic/pulsed/sequence_generator_logic.py
+++ b/src/qudi/logic/pulsed/sequence_generator_logic.py
@@ -1701,259 +1701,269 @@ class SequenceGeneratorLogic(LogicBase):
         TODO: Add parameters that are stored
         """
         # Get PulseBlockEnsemble from saved ensembles if string has been passed as argument
-        if isinstance(ensemble, str):
-            ensemble = self.get_ensemble(ensemble)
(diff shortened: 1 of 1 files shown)


TobiasSpohn (dismissed):

TobiasSpohn (approved):
I have tested it since the last commit on a confocal setup and I found no issues. Thanks for the improvement!

PR #198, opened 2025-09-30
Cursor movement and scanning implemented for the AMC300

florianfeuchtmayr:
## Description
Added closed loop movement in AMC300_stepper. 
Fixed main problems of the AMC300_ni_scanning_probe_interfuse. When moving the cursor or slider, the movement is not executed immidietly. But a timer is started, which is reseted if the cursor or slider moves again. When the timer is finished, one movement to the new target positon is executed by a closed loop movement.
A scan is handled by the _run_scan_worker which moves closed loop to each pixel and acquires a certain number samples. Unnecessary functions, classes or variables were removed. Comments were added.
When checking the constraints in the optimizer logic, the relative scan ranges are checked if the check of the absolute scan ranges fails.
Magnet files that were only there for testing are removed.

## Motivation and Context
The AMC300 tools were not usable. 

## How Has This Been Tested?
This was tested on the hardware with an AMC300 controller connected. Movement and scans were performed without errors.

## Types of changes

- [x] Bug fix
- [x] New feature

## Checklist:


florianfeuchtmayr:
Sorry,  was done by mistake

PR #204, opened 2025-10-27
Keysight external clock fix

TobiasSpohn:
## Description

Fixes setting the external clock on Keysight AWG's through the configuration file by first setting the frequency of the external clock and then activating the external reference clock.

## Motivation and Context

When using the ``ConfigOption``'s to activate the external clock on Keysight AWG's it can happen that when first activating the external clock and the set frequency does not match the supplied frequency that the ``Keysight Soft Panel`` software crashes and the module can't be activated. By first changing the set frequency to the correct one and then activating, the software does not crash. @prithviulm also had this problem on the dilution fridge setup.

## How Has This Been Tested?

Running on my confocal setup since the first commit.

## Types of changes

- [x] Bug fix
- [ ] New feature

## Checklist:

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -8,6 +8,7 @@
 
 ### Bugfixes
 - Fixed code suggestions for `Connector` objects
+- Fixed Keysight AWG's from breaking when setting the external reference clock through the configuration file
 
 ### New Features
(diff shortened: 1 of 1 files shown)


prithviulm:
Can confirm this was the fix I used to stop the Keysight software crashing.

PR #219, opened 2026-01-12
Automatic sequence count length determination

TobiasSpohn:
## Description

This PR introduces the `PredefinedGeneratorBase._get_sequence_count_length` method that can automatically determine the count length of a predefined `PulseSequence`.
It works similar to the  `PredefinedGeneratorBase._get_ensemble_count_length` method, adjusted to a `PulseSequence`.

## Motivation and Context

For now, each predefined method generated `PulseSequence` required the writer to set the counting length manually, while it was possible for `PulseBlockEnsemble`'s to do that automatically with the `PredefinedGeneratorBase._get_ensemble_count_length` method. It would ease the creation of `PulseSequence`'s and reduce code duplication if there was a similar function for a `PulseSequence`.

Furthermore, when using the `t1_sequencing` predefined method it creates the wrong counting length for a gated counter, by creating the same counting length as for the ungated counter, making the current implementation not usable with a gated counter.

## How Has This Been Tested?

confocal setup and dummy setup

## Types of changes

- [x] Bug fix
- [x] New feature

## Checklist:

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -11,6 +11,8 @@
 - Fixed code suggestions for `Connector` objects
 - Fixed Keysight AWG's from breaking when setting the external reference clock through the configuration file
 - Fixed ``laser_logic`` to work with remote laser hardware
+- Fixed typo in ``MicrowaveAnritsu.on_activate`` preventing the module from activating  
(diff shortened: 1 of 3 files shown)


takuya-ulm (dismissed):
Thanks.
I chacked proper counter record length for gated and ungated mode.

PR #222, opened 2026-01-15
Mw anritsu bugfix

PENG-WANG0601:
<Fixed typo in MW Anritsu source MG37022A hardware file>

## Description
Changed one small bug, where it tried to connect to gpib instead of the device, which in this case could be connected via USB.

## Motivation and Context

This solves the problem of connection as before it could only be connected via gpib, now it is also possible with USB.

## How Has This Been Tested?

Changes were tested with Anritsu MG37022A MW source, using the latest version of qudi main qudi-iqo-modules installed on a Windows 11 PC, typical confocal setup. Hardware connection was tested using the CW ODMR Gui & Logic, output was monitored using a spectrum analzyer (Anritsu MS2726C). Both, the static output as well as the sweep could be observed.

## Types of changes

- [x] Bug fix
- [ ] New feature

## Checklist:

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -11,7 +11,7 @@
 - Fixed code suggestions for `Connector` objects
 - Fixed Keysight AWG's from breaking when setting the external reference clock through the configuration file
 - Fixed ``laser_logic`` to work with remote laser hardware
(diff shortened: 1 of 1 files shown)


TobiasSpohn (approved):
Thanks for fixing

PR #224, opened 2026-02-05
Fixed Inheritance of Sampling Function Params

fab1an-m:
## Description

When one generates functions that inherit from one of the basic sampling functions, e.g. when creating shaped pulses, they should have both parameters from their own class as well as from the sampling function. Therefore they should inherit the `params` dict content of the basic sampling function as well as have their `params` dict content from their own class. To fix this, a instance `params` dict is introduced and updated with the class `params` in the `__init__()` method. This way any pulse object that inherits from one of the basic sampling functions has all parameters stored in its  `self.params` instance variable.

## Motivation and Context

Some sampling functions inherit from the basic sampling functions. The problem was, that the inheritance of the `params` variable didn't work when both classes have params variables and any change to `self.params` was changing the class variable `params` and therefore for all objects of the sampling function class. This has lead to errors in the pulsed measurement gui pulse generator when displaying the parameters of different pulse elements with different parameters that both inherit from the same basic sampling function.

## How Has This Been Tested?

Tested on confocal setup with normal and shaped pulses for e.g. Sine and TripleSinSum sampling function

## Types of changes

- `super.__init__()` call for all sampling function classes in the init method
- assignment of the `params` class variable to `self.params` in the init method of all sampling classes
- default assignment of an empty dict in `SamplingBase` `__init__()` method to ensure the instance variable `params` exists for all objects that inherit from `SamplingBase`
- When generating normal sampling functions, the only change is that there is now both the params instance variable as well as the params class variable. If you want to change the default params for a sampling function class, you have to change the class variable, not the instance variable of an object of this class. For an object you can just change `self.params` without having to worry that it changes the `params` for future objects.
- 

- [x] Bug fix
- [ ] New feature

## Checklist:

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -11,9 +11,10 @@
 - Fixed code suggestions for `Connector` objects
 - Fixed Keysight AWG's from breaking when setting the external reference clock through the configuration file
 - Fixed ``laser_logic`` to work with remote laser hardware
-- Fixed typo in ``MicrowaveAnritsu.on_activate`` preventing the module from activating  
+- Fixed typo in ``MicrowaveAnritsu.on_activate`` preventing the module from activating
 - Fixed `KeysightM8195A` not loading `PulseBlock`s in the pulsed main GUI
 - Fixed counting length for the `t1_sequencing` predefined generate method for a gated counter
+- Fixed sampling functions inheritance of parameters
 
 ### New Features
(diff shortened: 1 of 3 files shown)


TobiasSpohn (dismissed):
Thanks for the improvements!
I simplified the creation of the sine parameters a bit by introducing a new inheriting class. Before merging,  please verify that this commit works for you as well.

TobiasSpohn (approved):

PR #241, opened 2026-05-02
added displaying logs in the test workflow

mohitulm:
## Description
This PR introduces changes to the test workflow that make Qudi logs visible during the workflow run if the tests fail.

## Motivation and Context
During testing, the logs/print statements inside Qudi do not appear in pytest's console. So, in case of a test failure, debugging is difficult. Printing the Qudi log after test failure grants some visibilty. 

## How Has This Been Tested?
This has been tested on a fork

## Types of changes
- [ ] Bug fix
- [x] New feature

## Checklist:


TobiasSpohn (approved):
Thanks for adding the logs to failed tests

PR #244, opened 2026-06-29
fixed some pyside6 issues in POI manager

mohitulm:
### Fix two PySide6 migration issues in POI manager

Both bugs surfaced after the Qt5 → PySide6/Qt6 migration and made the POI
manager partially unusable.

### 1. Selecting a POI from the dropdown raises "POI name must be of type str or None"

`active_poi_ComboBox` was connected via `activated[int]`, which emits the
selected *index*, so `set_active_poi` received an integer instead of the POI
name. Qt6 removed the `activated(const QString&)` overload that the original
code(Pyside2) relied on.

Fixed by connecting to `textActivated`, the Qt6 replacement that emits the
selected text. Applied to both the connect and disconnect sites.

### 2. "New POI" no longer placed at the current scanner position

New POI can be added either by a mouse click event or defaulting to the current scanner position. Both ways , the method add_poi() is called in poi_manager_logic. In pyside2 with the default, the method was called with no parameters, but after the migration it gets called with a bool type and thus position is not fetched from the scanner. Adding a condition to check if the position is False fixes this. 

A similar issue was when clicking  Move POI anchor or Manual update Position buttons. This is also caused because a slot gets called with bool type.  Adding a condition to check if the name string is bool fixes this.

## How Has This Been Tested?
This has been tested locally and seems to work as before with pyside2

## Types of changes
- [x] Bug fix
- [ ] New feature

## Checklist:

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -6,6 +6,7 @@
 
 ### Bugfixes
 - Fix `installtion.py` for arbitrary Python versions
+- Fix several `POIManagerGUI` `PySide6` bugs
 
 ### New Features
 
--- a/src/qudi/gui/poimanager/poimanagergui.py
+++ b/src/qudi/gui/poimanager/poimanagergui.py
@@ -474,7 +474,7 @@ class PoiManagerGui(GuiBase):
             self._poi_manager_logic().set_poi_anchor_from_position, QtCore.Qt.ConnectionType.QueuedConnection)
         self._mw.delete_poi_PushButton.clicked.connect(
             lambda: self._poi_manager_logic().delete_poi(None), QtCore.Qt.ConnectionType.QueuedConnection)
-        self._mw.active_poi_ComboBox.activated[int].connect(
+        self._mw.active_poi_ComboBox.textActivated.connect(
             self._poi_manager_logic().set_active_poi, QtCore.Qt.ConnectionType.QueuedConnection)
         self._mw.goto_poi_after_update_checkBox.stateChanged.connect(
             self._poi_manager_logic().set_move_scanner_after_optimise, QtCore.Qt.ConnectionType.QueuedConnection)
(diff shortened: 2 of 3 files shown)


TobiasSpohn (changes requested):
Thanks for the PR. There are still some buttons not working

when adding a new POI and then clicking on `Move POI anchor` or `Manual update Position` there is this error:

```python
  File "/home/tobi/qudi/qudi-iqo-modules/src/qudi/logic/poi_manager_logic.py", line 845, in set_poi_anchor_from_position
    shift = position - self.get_poi_position(name)
                       ~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/home/tobi/qudi/qudi-iqo-modules/src/qudi/logic/poi_manager_logic.py", line 785, in get_poi_position
    return self._roi.get_poi_position(name)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/home/tobi/qudi/qudi-iqo-modules/src/qudi/logic/poi_manager_logic.py", line 189, in get_poi_position
    raise TypeError('POI name must be of type str.')
TypeError: POI name must be of type str.
```

In the text field above these 3 buttons there is `POI distance to cursor (rel_x, rel_y)`, there is always a (?, ?) displayed

When saving an ROI and loading it again, there is the Warning:
```python
`/home/tobi/qudi/qudi-iqo-modules/src/qudi/logic/poi_manager_logic.py:1236: UserWarning: Input line 1 contained no data and will not be counted towards `max_rows=50000`.  This differs from the behaviour in NumPy <=1.22 which counted lines rather than rows.  If desired, the previous behaviour can be achieved by using `itertools.islice`.
Please see the 1.23 release notes for an example on how to do this.  If you wish to ignore this warning, use `warnings.filterwarnings`.  This warning is expected to be removed in the future and is given only once per `loadtxt` call.
  poi_names = np.loadtxt(complete_path, delimiter='\t', usecols=0, dtype=str, ndmin=1)
```

Can you please add fixes to the above errors as well?

mohitulm:
Hi @TobiasSpohn , thanks for pointing out these issues. I have updated the PR to address the 'Move POI anchor' or 'Manual update Position' buttons. 

 (?, ?) is the default value for poi_distance_label in the UI file and label is not handled anywhere in the logic. 

As for the last warning, it doesn't affect the functionality and ROI can be correctly loaded.

TobiasSpohn (dismissed):
Thanks for the fixes

PR #246, opened 2026-07-07
supporting customized alt plot methods in pulsed analysis

mohitulm:
Make the analysis-tab alternative plot extensible via a base class

## Description
The "alternative plot" (second plot) in the pulsed analysis tab previously supported only
FFT and Delta, with the computation and axis labels hardcoded in `PulsedMeasurementLogic`
and the GUI dropdown items hardcoded in the `.ui`. This PR replaces that with a discovery-
based plugin architecture mirroring `PulseAnalyzer`/`PulseExtractor`: users add a method by
dropping a subclass of `AltPlotMethodBase` into a folder, and it automatically becomes
selectable in the GUI. FFT and Delta are reimplemented as the first two such methods.

Adding a new alternative plot method :
Subclass `AltPlotMethodBase`, set a unique `name`, and implement
`compute(self, signal_data, **kwargs)` returning a 2D array (row 0 = x-axis, rows 1.. = data)
or `None` to fall back to a flat plot. Any keyword arguments of `compute` (with correctly-
typed defaults) are auto-discovered as configurable, persisted parameters. Optionally override
`is_available()` (e.g. require an alternating measurement) and the `x_axis_*`/`y_axis_*` label
properties. The method then appears in the analysis-tab dropdown automatically.

Look at the existing FFT and Delta plots that have have been reimplemented in 'basic_alt_plot_methods.py'. Any new plots can be added similarly.

## How Has This Been Tested?
This has been tested locally with dummy hardware modules.

## Types of changes
- [ ] Bug fix
- [x] New feature

## Checklist:

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -14,6 +14,7 @@
 - Fix several `POIManagerGUI` `PySide6` bugs
 
 ### New Features
+- supporting custom alternative plots in pulsed analysis 
 - Added resizeable `PulseBlockEditor` columns in `pulsed_gui`
 - Added pulse shaping to the `pulsed` tool chain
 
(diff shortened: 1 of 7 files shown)


TobiasSpohn (approved):
Thanks for the improvement

PR #247, opened 2026-07-09
Feature: Pid Gui window title based on module name

JeppeSparrow:
## Description
Makes the window title of the pid gui follow the name of the module it's loaded from.

## Motivation and Context
Makes it easier to distinguish windows if there is more than one pid gui present.

## How Has This Been Tested?
Tested with `default.cfg`, `just build` recipe passes.

## Types of changes

- [ ] Bug fix
- [x] New feature

## Checklist:


JeppeSparrow:
Wrong target. Apologies!

PR #250, opened 2026-08-03
fix ni io samples_in_buffer

qku:
## Description

Check the length of `_di_task_handles` instead of checking if it is `None`.

## Motivation and Context

I ran into this issue when I tried to configure only an analog input and no digital inputs. The method samples_in_buffer differentiates betwen three cases:

- no analog inputs, only digital inputs
- only analog inputs, no digital inputs
- both analog and digital inputs

to correctly compute the number of samples available in the buffer. If no digital inputs are configured, `_di_task_handles` is an empty list, and not `None`. Therefore, the checking fails and the third case is assumed incorrectly, which leads to an error.

## How Has This Been Tested?

Configured the NI IO with only digital, only analog and both analog and digital channels.

## Types of changes

- [x] Bug fix
- [ ] New feature

## Checklist:

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -13,6 +13,7 @@
 - Fixed `sequence_generator_logic` configuration in `default.cfg`
 - Fix several `POIManagerGUI` `PySide6` bugs
 - Fix discovery of `AltPlotMethodBase` subclasses
+- Fix NI IO for only analog input channels.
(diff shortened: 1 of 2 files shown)


TobiasSpohn (approved):
Thank you very much for the bugfix.

PR #252, opened 2026-08-17
fast_counter_dummy Improvement

Smitsavaliya49:
Fast-counter dummy
Fixed resolution of the default demo trace path.
Count data is initialised during module activation.
Stored and normalized the configured number of gates.
Added emulation of configured hardware binning by summing base-clock bins.
Added cropping or repetition of demo trace data to match the configured record length.

Added correct two-dimensional output for gated measurements with shape:

(number_of_gates, number_of_time_bins)

Ensured ungated measurements return a one-dimensional trace of the configured length.
Improved trace-loading error handling and logging.
Set the counter to the error state when the trace cannot be loaded.
Changed the counter to the running state only after trace preparation succeeds.

### DIFF
--- a/src/qudi/hardware/dummy/fast_counter_dummy.py
+++ b/src/qudi/hardware/dummy/fast_counter_dummy.py
@@ -20,10 +20,10 @@ You should have received a copy of the GNU Lesser General Public License along w
 If not, see <https://www.gnu.org/licenses/>.
 """
 
-import os
 import time
-
+import os
 import numpy as np
+from scipy import ndimage
 
 from qudi.core.configoption import ConfigOption
 from qudi.interface.fast_counter_interface import FastCounterInterface
@@ -45,6 +45,17 @@ class FastCounterDummy(FastCounterInterface):
     # config option
     _gated = ConfigOption('gated', False, missing='warn')
     trace_path = ConfigOption('load_trace', None)
+    # Optional manual override for the number of laser pulses to simulate in ungated
+    # mode. Normally this is discovered automatically from the running measurement
+    # (see `_number_of_points`), so it only needs to be set when running the dummy
+    # without a PulsedMeasurementLogic.
+    _ungated_points = ConfigOption('ungated_points', None)
+
+    # The demo trace file was recorded for a Rabi measurement with exactly this many
+    # laser pulses/points. Its individual laser pulses are located by flank detection
+    # and re-emitted at a regular period, so that any requested number of points can
+    # be built by cycling through them.
+    _reference_points = 50
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
@@ -62,8 +73,8 @@ class FastCounterDummy(FastCounterInterface):
         self.statusvar = 0
         self._binwidth = 1
         self._gate_length_bins = 8192
-        self._number_of_gates = 1 if self._gated else 0
-        self._count_data = np.zeros(self._gate_length_bins, dtype='int64')
+        self._number_of_gates = 0
+        self._points_origin = 'gate count'
         return
 
     def on_deactivate(self):
@@ -73,10 +84,10 @@ class FastCounterDummy(FastCounterInterface):
         return
 
     def get_constraints(self):
-        """ Retrieve the hardware constraints from the Fast counting device.
+        """ Retrieve the hardware constrains from the Fast counting device.
 
         @return dict: dict with keys being the constraint names as string and
-                      items are the definition for the constraints.
+                      items are the definition for the constaints.
 
          The keys of the returned dictionary are the str name for the constraints
         (which are set in this method).
@@ -110,16 +121,11 @@ class FastCounterDummy(FastCounterInterface):
 
         # the unit of those entries are seconds per bin. In order to get the
         # current binwidth in seconds use the get_binwidth method.
-        constraints['hardware_binwidth_list'] = [
-            1/950e6,
-            2/950e6,
-            4/950e6,
-            8/950e6
-        ]
+        constraints['hardware_binwidth_list'] = [1/950e6, 2/950e6, 4/950e6, 8/950e6]
 
         return constraints
 
-    def configure(self, bin_width_s, record_length_s, number_of_gates=0):
+    def configure(self, bin_width_s, record_length_s, number_of_gates = 0):
         """ Configuration of the fast counter.
 
         @param float bin_width_s: Length of a single time bin in the time trace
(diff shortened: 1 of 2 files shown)


src/qudi/hardware/dummy/fast_counter_dummy.py
```diff
@@ -20,8 +20,9 @@
 If not, see <https://www.gnu.org/licenses/>.
 """
 
-import time
 import os
+import time
+
```
TobiasSpohn: the order of the imports does not need to change

src/qudi/hardware/dummy/fast_counter_dummy.py
```diff
@@ -105,12 +109,17 @@ def get_constraints(self):
...
         # the unit of those entries are seconds per bin. In order to get the
-        # current binwidth in seonds use the get_binwidth method.
-        constraints['hardware_binwidth_list'] = [1/950e6, 2/950e6, 4/950e6, 8/950e6]
+        # current binwidth in seconds use the get_binwidth method.
+        constraints['hardware_binwidth_list'] = [
+            1/950e6,
+            2/950e6,
+            4/950e6,
+            8/950e6
+        ]
```
TobiasSpohn: This formatting does not need to change

src/qudi/hardware/dummy/fast_counter_dummy.py
```diff
@@ -147,16 +159,64 @@ def get_status(self):
...
+
+        usable_size = count_data.size - (count_data.size % self._binwidth)
+        if usable_size == 0:
+            return np.zeros(0, dtype='int64')
+        return (
+            count_data[:usable_size]
+            .reshape(-1, self._binwidth)
+            .sum(axis=1)
+            .astype('int64')
+        )
```
TobiasSpohn: why do we need this function?

src/qudi/hardware/dummy/fast_counter_dummy.py
```diff
@@ -147,16 +159,64 @@ def get_status(self):
...
+            return np.zeros(target_size, dtype='int64')
+        if count_data.size >= target_size:
+            return count_data[:target_size].astype('int64', copy=False)
+
+        repetitions = int(np.ceil(target_size / count_data.size))
+        return np.tile(count_data, repetitions)[:target_size].astype(
+            'int64',
+            copy=False
+        )
+
```
TobiasSpohn: Rename this function to a more easy to understand name, e.g. `_fit_count_data_to_size`

src/qudi/hardware/dummy/fast_counter_dummy.py
```diff
@@ -206,7 +266,8 @@ def get_data_trace(self):
         The binning, specified by calling configure() in forehand, must be
         taken care of in this hardware class. A possible overflow of the
         histogram bins must be caught here and taken care of.
-        If the counter is NOT GATED it will return a tuple (1D-numpy-array, info_dict) with
+        If the counter is NOT GATED it will return a tuple
+        (1D-numpy-array, info_dict) with
```
TobiasSpohn: formatting does not need to change

TobiasSpohn (changes requested):
Thanks for looking into this. For me however, the fixes do not work on a dummy setup. Please test it again on your end, the easiest way is to load a Rabi measurement with 50 data points, that is how this data was taken. The result after analysis should be a Sine oscillation.

With these changes this no longer yields an oscillation for me in ungated mode. Furthermore, when adding more points in e.g. the Rabi measurement the data should just be cycled to match the new number of points. 
The same is true for me with the gated mode, there I do no longer see an oscillation.

It seems like something with the data reshaping is not working properly. You can use the main branch and generate the Rabi sequence there, the result in ungated yields an oscillation with 50 points and in the pulse extraction tab the laser pulses are cleanly visible. This is how the result should look like with the fixes.

Furthermore, please add an entry to the changelog.

TobiasSpohn:
Thanks for the changes now gated and ungated works as expected. I noticed that in gated mode the length of the individual laser pulses is a lot longer than the actual data when using the pass through extraction method. Probably something with the calculation of the lengths of each data pulse is still going wrong.

image

src/qudi/hardware/dummy/fast_counter_dummy.py
```diff
@@ -44,14 +45,26 @@ class FastCounterDummy(FastCounterInterface):
...
+    # mode. Normally this is discovered automatically from the running measurement
+    # (see `_number_of_points`), so it only needs to be set when running the dummy
+    # without a PulsedMeasurementLogic.
+    _ungated_points = ConfigOption('ungated_points', None)
+
+    # The demo trace file was recorded for a Rabi measurement with exactly this many
+    # laser pulses/points. Its individual laser pulses are located by flank detection
+    # and re-emitted at a regular period, so that any requested number of points can
+    # be built by cycling through them.
+    _reference_points = 50
```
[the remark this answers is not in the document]
TobiasSpohn: we want to introduce as little as possible changes while enabling the use of gated counting and the ability to select an arbitrary number of laser pulses

src/qudi/hardware/dummy/fast_counter_dummy.py
```diff
@@ -60,6 +73,8 @@ def on_activate(self):
         self.statusvar = 0
         self._binwidth = 1
         self._gate_length_bins = 8192
+        self._number_of_gates = 0
+        self._points_origin = 'gate count'
```
TobiasSpohn: for what is this needed? this should be inferred if the dummy is in gated or ungated mode configured, no need for an additional variable

Smitsavaliya49:
> Thanks for the changes now gated and ungated works as expected. I noticed that in gated mode the length of the individual laser pulses is a lot longer than the actual data when using the pass through extraction method. Probably something with the calculation of the lengths of each data pulse is still going wrong.
> 
> image

Thanks for checking! I think the pulse lengths are acting up because the gate_channel isn't selected. In the Pulse Generator's tab, try setting the gate_channel to d_ch2 instead of leaving it empty. That solved it on my end! Let me know if it still doesn't work for you after setting it.

TobiasSpohn (changes requested):
I get that it's not easy to reshape/expand this sample data properly. However, let's not introduce a lot of helper code to work with this example data. We can also just simulate the data following a sine and construct a simulated laser pulse from it. We have done this already in PR #110. Let's do something similar here.  

 If it is this difficult to calculate the ungated mode data correctly, let's just keep it like it is currently, so it will always output the 50 laser pulse trace with the default laser length of 3 us and a laser delay of 500 ns and give a warning that the data can only always be max 50 laser pulses. To make the conv_deriv extraction method working let's introduce spacings with zeros of the same length between the laser pulses.

For gated the simulation will be straight forward, just sample the sine at number_of_gates points with record_length_s / bin_width_s number of bins.

We can do it similar to #110 that we have just rectangular laser pulses or we can simulate similar behavior to the real laser pulse that there will be a first rectangle that gives the state information, so it contains the data of the sine and then a second rectangle that fills up to the laser length which is just a constant base line. I would opt for the second variant, then the analysis and norm window become more clear. 

Can you put these helper functions that create the simulated data into a separate file in the dummy folder, then we can use it for QDyne as well and have a common data simulation.

TobiasSpohn:
> Thanks for checking! I think the pulse lengths are acting up because the gate_channel isn't selected. In the Pulse Generator's tab, try setting the gate_channel to d_ch2 instead of leaving it empty. That solved it on my end! Let me know if it still doesn't work for you after setting it.

It did indeed improve it, however the data is still too long, but maybe this is because it is just cut out from the raw data trace. I think with simulated data this issue would be solved.

TobiasSpohn (approved):
Thanks for working on this improvement of the counter dummy

